PR #4, opened 2022-05-31
Initial version of EESSI bot using PyGHee

trz42:
bot now uses PyGHee. It handles a few events (by just printing log
information). Most changes are done for building a software stack upon
receiving a pull request with action opened.

Major changes:
- expanded app.cfg.example file containing configuration settings for
  the bot, values need to be adjusted to where the bot runs and made
  available as app.cfg in the directory from where the bot runs.
- Largely reworked function build_easystack_from_pr defined in
  tasks/build.py
  - makes use of extended app.cfg
  - contains lots of logging info (could be removed later on or wrapped
    in some debug level)
  - creates a separate job dir for each delivery of an event (good for
    debugging during development), each EESSI version and each
    (micro)architecture
  - applies some whitelist to limit for which EESSI version it should
    build
  - makes use of app specific definition of target build architectures
    and needed job parameters to constrain job to a node with that
    architecture

Medium changes:
- Adjusted eessi_bot_software_layer.py to be based on PyGHee. Function
  read_event_from_json moved to PyGHee.

Minor changes:
- Deleted handlers directory. Handlers now directly defined in class
  EESSIBotSoftwareLayer.
- New way to run bot (in run.sh).


scripts/build_with_EESSI_install.slurm
```diff
@@ -0,0 +1,95 @@
...
+
+# the container script launches the container
+#${SCRIPTS_DIR}/launch_build_container.sh ${SCRIPTS_DIR}/run_in_compat_layer_env.sh ${SCRIPTS_DIR}/mod_EESSI_install.sh
+export EESSI_PILOT_VERSION=${BUILD_EESSI_VERSION}
+singularity exec \
+    --bind "${SCRIPTS_DIR}:${SCRIPTS_DIR}" \
+    --fusemount "${EESSI_PILOT_READONLY}" \
+    --fusemount "${EESSI_PILOT_WRITABLE_OVERLAY}" \
+    docker://ghcr.io/eessi/build-node:debian10 \
+    ${SCRIPTS_DIR}/run_in_compat_layer_env.sh ${SCRIPTS_DIR}/mod_EESSI_install.sh
```
boegel, line 94: You should be able to remove this whole block, including defining the related environment variables above, with:

```
export SINGULARITY_BIND="${SCRIPTS_DIR}:${SCRIPTS_DIR}"
./build_container.sh run ${LOCAL_TMP}/${USER}/EESSI ${SCRIPTS_DIR}/mod_EESSI_install.sh
```

That's using the scripts available in https://github.com/EESSI/software-layer/

One small change is needed for `build_container.sh`, since it hard defines `$SINGULARITY_BIND` currently, without taking into account an existing value.

So this line https://github.com/EESSI/software-layer/blob/7ec49a2885cd5e01d808ce6e753711d748f1234a/build_container.sh#L44

should be tweaked to something like:

```shell
# take into account that $SINGULARITY_BIND may be defined already
BIND_PATHS="$EESSI_TMPDIR/var-run-cvmfs:/var/run/cvmfs,$EESSI_TMPDIR/var-lib-cvmfs:/var/lib/cvmfs,$EESSI_TMPDIR"
if [ -z $SINGULARITY_BIND ]; then
    export SINGULARITY_BIND="$BIND_PATHS"
else
    export SINGULARITY_BIND="$SINGULARITY_BIND:$BIND_PATHS"
fi

File: scripts/build_with_EESSI_install.slurm [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+#   scripts dir could be a directory known through app.cfg or stored in job definition file
+echo "export SCRIPTS_DIR=${SCRIPTS_DIR}" >> _env
+#   only used to set EESSI_TMPDIR, could be taken from a job definition file
+echo "export LOCAL_TMP=${LOCAL_TMP}" >> _env
+#   probably not needed (because all information can be derived via scripts on build node, i.e.,
+#   inside a job); same as "${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR}"
+echo "export BUILD_TARGET_ARCHITECTURE=${BUILD_TARGET_ARCHITECTURE}" >> _env
+#   could be delivered as argument, only used to set EESSI_PILOT_VERSION, the latter might be
+#   set directly instead
+echo "export BUILD_EESSI_VERSION=${BUILD_EESSI_VERSION}" >> _env
```
boegel, line 84: Why not just pass these variables as arguments to the `mod_EESSI_install.sh` script instead?

Maybe leave it as is for now, since I suspect most of this will no longer be required if we change tactics to just run an `eessi-bot-build.sh` script from the software-layer...

boegel:
@trz42 For now, I would also add `mod_EESSI_install.sh` to this PR, so we have a complete setup in the repo of the working bot (even if we'll change tactics in the next PR).

`mod_EESSI_install.sh` is a copy of the `EESSI-pilot-install-software.sh` in the software-layer repo which was "hacked" to only install stuff that is specified in a specific easystack file, which is good enough by means of a proof-of-concept.

Maybe remote the commented out parts, just to make the script smaller/easier to digest...

eessi_bot_software_layer.py
```diff
@@ -9,65 +9,81 @@
 #
```
boegel, line 9: @trz42 Add yourself as an author to this script (and also add @bedroge while you're at it)

eessi_bot_software_layer.py
```diff
@@ -9,65 +9,81 @@
...
+
+    def handle_pr_opened_event(self, event_info, pr):
+        """
+        Handle opening of a pull request.
+        """
+        log("PR opened")
+        build_easystack_from_pr(pr, event_info)
+
+
+    def handle_pull_request_event(self, event_info, log_file=None):
```
boegel, line 62: This looks like a (default) generic handler that could be moved into PyGHee, no?

It basically refines a PR event to something more fine-grained using the `action` field (PR opened, PR labeled, etc.)

To avoid confusion, I would name the methods triggered by this `handle_pull_request_<action>_event` (rather than using `_pr_`).

You can also avoid the `handlers` dict by using something like this:

```python
handler_name = 'handle_pull_request_%s_event' % action
if hasattr(self, handler_name):
    handler = getattr(self, handler_name)
    handler(event_info, pr)
else:
    log("No handler for PR action '%s'" % action)
```

For now, we can leave this as is, and clean it up later once PyGHee has been enhanced with a method to refine a `pull_request` event into more specific handlers.

eessi_bot_software_layer.py
```diff
@@ -78,15 +94,15 @@ def main():
...
     elif opts.cron:
         log("Running in cron mode")
     else:
         # Run as web app
-        app = create_app()
-        app.run()
-
+        app = create_app(klass=EESSIBotSoftwareLayer)
+        log("EESSI bot for software layer started!")
+        waitress.serve(app, listen='*:3000')
```
boegel, line 105: This is also something small we can move to PyGHee: this should just become `app.start(port=3000)` or something like that (which avoids having to import `waitress` here, that should be internal to PyGHee)

tasks/build.py
```diff
@@ -1,30 +1,113 @@
...
+
+    # [architecturetargets]
+    arch_target_map = json.loads(config.get_section('architecturetargets').get('arch_target_map'))
+    log("arch target map '%s'" % (json.dumps(arch_target_map)))
+
+    # get event id and create main directory for event and a temporary dir
+    # inside that (the latter uses mktemp to create a new job every time the
+    # event is being redelivered, i.e., useful while developing)
+    event_id = event_info['id']
+    jobs_dir = os.path.join(jobs_base_dir, event_id)
```
boegel, line 41: There should be a top directory for this specific PR here to give this some more structure, i.e. `pr<id>/jobs/<event_id>`, ...

Maybe we also need another directory level on top, per month (`2022.06/pr<id>/...`), so we can easily tell which old stuff we can remove or archive, and to avoid having lots of subdirectories

Not terribly important for now, can be changed later.

tasks/build.py
```diff
@@ -1,30 +1,113 @@
...
+
+    # - determine files to proceed and prepare job dirs (essentially we
+    #   may have gotten multiple easystack files (N) via a single PR and
+    #   the bot may be configured to build for different
+    #   (micro)architectures (M); resulting in N x M jobs)
+    # - the N files may be filtered by white listed versions (bot config)
+    # - we assume that the full path to the easystack file includes
+    #   an EESSI version, e.g., 2021.12/softwarelist.yaml
+    jobs = []
+    for pr_file in repo.get_pull(pr.number).get_files():
```
boegel, line 63: We don't want to download individual files eventually, we just want to get a checkout of the entire PR branch, and use that as a working directory for the job.

With the current approach of just "reacting" to specific easystack files, this is fine, but it can probably be cleaned up quite a bit...

tasks/build.py
```diff
@@ -1,30 +1,113 @@
...
+        d["SCRIPTS_DIR"] = scripts_dir
+        d["LOCAL_TMP"] = local_tmp
+        d["BUILD_TARGET_ARCHITECTURE"] = job[1]
+        d["BUILD_EESSI_VERSION"] = job[0]
+        log("Submit job with '%s %s %s/%s' from directory '%s'" % (submit_command, job[2], scripts_dir, build_job_script, job[3]))
+        # alternative to command_line as string would be a list of strings each without a whitespace, e.g.,
+        # [submit_command, job[2] /*splitted*/, scripts_dir + '/' + build_job_script]
+        command_line = submit_command + ' ' + job[2] + ' ' + scripts_dir + '/' + build_job_script;
+        submitted_job = subprocess.run(command_line, shell=True, cwd=job[3], env=d, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        log("Submit command executed!\nStdout: %s\nStderr: %s" % (submitted_job.stdout, submitted_job.stderr))
```
boegel, line 113: Future work: also check exit code, so we can deal with problems with job submission, etc.

boegel (approved):

scripts/build_with_EESSI_install.slurm
trz42: Originally thought that I could just set these as environment variables by the bot before submitting the job. Only later I figured out that running in the combat layer cleans up all environment settings.

Agree to leave it and revise once we have the more generic approach (via `eessi-bot-build.sh`) in place.

eessi_bot_software_layer.py
trz42: Just changed this as suggested.

tasks/build.py
trz42: Created issue #5 to take care of this.

tasks/build.py
trz42: Created issue #6 as a reminder for this.

scripts/build_with_EESSI_install.slurm
trz42: Created issue #8 to keep this in mind.

eessi_bot_software_layer.py
trz42: Created issue #9 for this.

boegel:
@trz42 Thanks for opening those issues to follow up on the suggestions!

I'll close this PR since all commits included here are also included in #10, which has additional changes on top.
Let's follow up in there...

Issue #8, opened 2022-06-02
Improve way information is handed from bot to build job


trz42:
Currently some information is conveyed via environment variables which are added to executing `sbatch` via a dictionary (parameter to `subprocess.run`), then in the build script they are checked for existence and stored in the file `_env` in the job directory. The latter is necessary to hand them to the actual build script as this runs in the compat layer (script `run_in_compat_layer.sh` which clears the environment).

This looks very complicated. We should check which variables we actually need (and in which script), then try to convey the information in a better/easier way.

trz42:
Fixed by #10

Issue #7, opened 2022-06-02
Improve directory structure for processing pull requests


trz42:
Current structure is as follows

`event['X-GitHub-Delivery']/run-id/EESSI_PILOT_VERSION/OS_TYPE_ARCH_TARGET/stackfile`

where _run-id_ is randomised directory name generated with mkdtemp (e.g., `tmpqad1fuqj`), _EESSI_PILOT_VERSION_ is the EESSI pilot version for which software shall be built (e.g., 2021.12), _OS_TYPE_ARCH_TARGET_ is a concatenation of `EESSI_OS_TYPE` and `EESSI_SOFTWARE_SUBDIR` with '/' replaced by '_' (e.g., linux_x86_64_intel_haswell).

In the future the structure could be as follows

`YYYY.MM/pr<id>/jobs/<event_id>/<run_number>/<cpu_target>/<branch of pull request>` and
`YYYY.MM/pr<id>/tars/<event_id>/<run_number>/<cpu_target>`

(alternatively structure could be simply
`YYYY.MM/pr<id>/<event_id>/<run_number>/<cpu_target>/<branch of pull request>`)

with the following intentions

- `YYYY.MM` representing the year and month when this event was processed (could be used to later more easily cleanup outdated event process handlings)
- pr<id> just the id of the pull request, so can be easier investigated
- jobs -- all jobs for this pull request
- tars -- all artifacts being built and possibly to be transferred to Stratum 0 server
- <event_id> id of the event (event[X-GitHub-Delivery] in the version represented by #4)
- <run_number> a number representing the delivery of the event (particularly during development each event could be delivered multiple times) -- in the version in #4 this is a string generated by `mkdtemp` (where the order may be less obvious)
- <cpu_target> is a name representing the SOFTWARE SUBDIR, e.g., `x86_64_amd_zen2`('/' replaced by '_') -- might not be necessary for `tars` tree if tarballs' name include the cpu target/software subdirectory
- <branch of pull request> is the full content of the branch of the pull request

trz42:
Implemented by #10

Issue #5, opened 2022-06-02
Make handling of pull requests more generic


trz42:
Currently, the bot (PR#4) makes a number of assumptions about the pull request (updates to easy stack files, where these are stored, etc.) and how it is handled (requiring essentially a new script to build & install software).

In the future, the bot should simply download the branch of the pull request (for the software-layer repository, this branch by design includes nearly everything necessary for handling the pull request: scripts, configuration, easystack files, etc) and submit a job that only follows the instructions for manually building an EESSI software stack (https://eessi.github.io/docs/software_layer/build_nodes/). By using existing helper scripts of the software-layer the job script, likely part of [EESSI/eessi-bot-software-layer](https://github.com/EESSI/eessi-bot-software-layer) repository, can be very generic not requiring much knowledge of the internals of the [EESSI/software-layer](https://github.com/EESSI/software-layer) repository.

trz42:
Implemented by #10

Issue #9, opened 2022-06-02
Improve start of app (in PyGHee)


trz42:
Currently (PR #4), the following (lines 103-105) is done to start the app

```
app = create_app(klass=EESSIBotSoftwareLayer)
log("EESSI bot for software layer started!")
waitress.serve(app, listen='*:3000')
```

Ideally this should be largely moved to [PyGHee](https://github.com/boegel/PyGHee) which would allow us to move the dependency on `waitress` to PyGHee and then just start the app by something like

`app.start(port=3000)`

boegel:
Also useful would be to print something when the bot was started successfully.
This message should include the location of the log file that the bot is using...

PR #10, opened 2022-06-03
initial (more generic) version of EESSI bot using PyGHee

trz42:
This makes the bot more generic. It adds a new batch script `eessi-bot-build.slurm`. The bot's handling of a pull request has been changed accordingly. In particular the following issues are solved by this PR:
- #5 -- the bot now makes only very few assumptions, ie, it reacts to PRs on the software-layer repo, it downloads the full branch of the PR and then uses the existing scripts to build and EESSI stack (that is it does not analyse the PR in any way),
- #7 -- directory structure for now is as follows `/some_base_dir/YYYY.MM/pr_<id>/event_<id>/run_<number>/<cpu_target>/` and all input (branch of pull request) and results of the job (slurm output file and tarball) are placed into this directory, 
- #8 -- completely eliminated need for environment variables; only information handed over is the location of the tmp directory for the build container

Some unneeded files and configuration settings were removed. Authors and license where added to the batch script.

edit (by @boegel): requires:
* https://github.com/boegel/PyGHee/pull/3


eessi_bot_software_layer.py
```diff
@@ -78,15 +89,15 @@ def main():
...
         # Run as web app
-        app = create_app()
-        app.run()
-
+        app = create_app(klass=EESSIBotSoftwareLayer)
+        log("EESSI bot for software layer started!")
+        waitress.serve(app, listen='*:3000')
 
 if __name__ == '__main__':
     main()
```
boegel, line 103: Starting the bot currently doesn't produce any output at all, ideally it should print something to signal that the bot has been started successfully.
A useful message would be to print the location to the log, for example (which should probably be taken care of by the `start` method of the instance that is produced by PyGHee's `create_app` (so it can be fixed as a part of #9)

boegel (changes requested):
@trz42 I went through this in detail, and also tested it, this is looking great!

While going through it I couldn't resist to make a couple of minor tweaks left and right, but mostly regarding code style issues, see https://github.com/trz42/eessi-bot-software-layer/pull/1.

The basic functionality of this is spot on now imho, and it is an excellent basis to keep making progress on in subsequent PRs...

scripts/eessi-bot-build.slurm
```diff
@@ -0,0 +1,51 @@
...
+eessi_tmpdir=$1
+
+# map the current directory (job environment which contains checkout
+#   of the branch of the pull request) to a specific path inside the container
+#export SINGULARITY_BIND="$(pwd):/eessi_bot_job"
+
+# run standard EESSI software-layer build script via compat layer
+#   inside build container
+export SINGULARITY_HOME="$(pwd):/eessi_bot_job"
+./build_container.sh run ${eessi_tmpdir} ./install_software_layer.sh
```
boegel, line 40: This should make become a simple call to a `bot-build.sh` script that also sits in the `software-layer` repo (something to follow up on later)

scripts/eessi-bot-build.slurm
```diff
@@ -0,0 +1,51 @@
...
+# create tarball for the above build and place it in some specific
+#   directory on shared disc (just in main directory of job)
+#   see https://github.com/EESSI/eessi-bot-software-layer/issues/7
+EESSI_SILENT=1 source init/eessi_environment_variables
+software_subdir=$(echo ${EESSI_SOFTWARE_SUBDIR} | tr '/' '-')
+timestamp=$(date +%s)
+export TGZ=$(printf "eessi-%s-software-%s-%s-%d.tar.gz" ${EESSI_PILOT_VERSION} ${EESSI_OS_TYPE} ${software_subdir} ${timestamp})
+./build_container.sh run ${eessi_tmpdir} \
+    ./create_tarball.sh ${eessi_tmpdir} ${EESSI_PILOT_VERSION} \
+    ${EESSI_SOFTWARE_SUBDIR} /eessi_bot_job/${TGZ}
```
boegel, line 51: Same here, this can become significantly shorter, I think, by creating a `bot-create-tarball.sh` script that is just being run here (and fed a couple of input variables).

scripts/eessi-bot-build.slurm
trz42: Made issue #15 for to remember this.

scripts/eessi-bot-build.slurm
trz42: Created issue #16 to remember this.

trz42: Replaced by issue https://github.com/EESSI/software-layer/issues/180

scripts/eessi-bot-build.slurm
trz42: Replaced by issue https://github.com/EESSI/software-layer/issues/181

boegel (approved):

Issue #20, opened 2022-06-15
Retry communication with GitHub if first attempt failed


trz42:
Communicating with GitHub (e.g., adding a comment to a pull request) might fail for various reasons. For example, we once received a `401` response with the message `Bad credentials`. Then, retrying by restarting the bot AND redelivering the same event that eventually led to the comment action resulted in a successful operation.

Instead of just failing the bot could after waiting some time try to repeat the communication with GitHub. Only when it fails again, or repeatedly fails, it could stop trying and report the problem, e.g., via email.

trz42:
#21 addresses only the `Bad credentials` issue. There may be other communication problems to which the bot should react more flexibly, i.e., by retrying, by reporting, ...

trz42:
See also #30.

trz42:
This has been implemented for several calls to GitHub.

PR #21, opened 2022-06-16
Fixing renewal of expired access token

trz42:
Code was prepared to renew the token which expires normally after an hour. There was a slight mismatch in naming the variable holding the token: `token` vs `_token`.

This PR address one scenario of #20 (the issue should be kept open because there may be other scenarios in which the bot should tolerate communication failures).


boegel (approved):
lgtm

PR #22, opened 2022-06-16
Add port as option

trz42:
Small update to let one specify a port number on which the bot listens for connections/events. Smee client also has to use that port.

Specifying the port number can be necessary if several bots or bot instances run on the same node.


trz42:
This is quite small and included (even with a slight update) in #24.
Hence I close this one.

Issue #23, opened 2022-06-16
Monitor potentially existing jobs after bot is (re)started


trz42:
When the bot (re)starts, for example, after a crash, it should resume monitoring potentially existing jobs.

To implement this,
- the bot probably need to persist information about jobs on disk (after they have been submitted),
- this job handling likely need to be done in a separate thread such that the bot could still receive and handle new events.

Maybe the bot could do all monitoring of jobs (and processing of finished jobs) in a separate thread?

trz42:
PR #24 addresses this by splitting the bot in two processes (instead of a separate thread).

Issue #17, opened 2022-06-15
Change method for fetching PR


trz42:
In PR #10 the contents of a pull request for EESSI/software-layer are obtained in `tasks/build.py` with the code

```
        git_clone = 'git clone https://github.com/%s --branch %s --single-branch %s' % (repo_name, branch_name, arch_job_dir)
```

It was pointed out by @boegel that this may not be correct, because the build than may not include more recent upstream commits.

A better (or the only correct) way would be:
1. Clone the base branch (target of the pull request), _and_
2. Apply the changes of the pull request on top.

Step 1 could be a simple `git clone https://github.com/REPO_NAME` where `REPO_NAME` is taken from the `base` of the PR (in PR #10 it's taken from the `head`).

Step 2 could obtain the changes by accessing `https://github.com/REPO_NAME/pull/PR_NUMBER.patch` and then run `patch` `REPO_NAME` again would be taken from the `base` of the PR with the number `PR_NUMBER`.

~~One should investigate if PyGitHub provides functionality for executing some of these steps (instead of running `git` commands).~~ Seems PyGitHub is not meant for such operations. Hence, resorting to just running the following commands from `destination_directory` might do the "trick":

```
git clone https://github.com/REPO_NAME destination_directory
curl -L https://github.com/REPO_NAME/pull/PR.patch > PR.patch
git am PR.patch
```

Notes, `destination_directory` must be non-existing or empty or `git clone` fails; `-L` instructs `curl` to follow redirects.

trz42:
PR #24 implements the outlined procedure. It is refined by inserting a step after cloning the repository. The complete procedure looks as follows:

1. `git clone https://github.com/REPO_NAME arch_job_dir`
2. `cd arch_job_dir`
3. `git checkout PR-BASE-BRANCH`
4. `curl -L https://github.com/REPO_NAME/pull/PR_NUMBER.patch > PR_NUMBER.patch`
5. `git am PR_NUMBER.patch`

Step 3 allows to maintain individual branches for each version of the software layer.

Issue #13, opened 2022-06-03
Implement adding a comment to a PR


trz42:
The bot should be able to add comments to a PR, e.g., when a job has ended or the tarball has been generated.

trz42:
This is implemented in PR #24. As it might be used for other tools it could be good to generalise this - identify comment and apply change - and put it into its own function/module.

trz42:
Generalising is addressed (the identification part) in #32.

Issue #12, opened 2022-06-03
Implement bot waiting for end of batch jobs


trz42:
The bot should detect when a batch job ends / has ended.

trz42:
It seems there are a number of alternative options to implement this functionality:

1. After the bot submitted a job, it continues checking periodically if the job has finished. Pros: easy to implement; Cons: requires additional measures in case the bot crashes and it may add unnecessary load on the machine where the bot runs (plus repeatedly sending requests to Slurm)
2. An independent cronjob periodically checks the status of jobs. Pros: more efficient checking (just need to query Slurm once per time interval, then checks what the status of jobs is) and might keep the handling of requests more simple (not requiring larger number of threads/workers for handling webhooks); Cons: requires storing of list of job ids on disk, also needs a way to send response to GitHub, e.g., adding a comment to Pull Request, which might require additional information to be stored
3. Bot receives recurring event "check_job_status" to check job status. Pros: does not require an independent component (e.g., cronjob as in alternative 2), can check status efficiently for all jobs; Cons: requires a means to send recurring events (which is just another dependency ... to be maintained), requires storing of list of job ids on disk, also needs a way to send response to GitHub, e.g., adding a comment to Pull Request, which might require additional information to be stored

While maybe not the most elegant, it might be sufficient to implement option 1 as first step.

trz42:
PR#24 implements option 2 though not as cronjob.

PR #24, opened 2022-06-16
Check build job status and comment on PR

trz42:
This is a larger pull request. Among a few smaller changes (e.g., adding ability to specify port on which the bot listens, makes PR #22 obsolete, sorry) it adds functionality to monitor the status of jobs, then checks the result of finished jobs and updates the PR accordingly.

The monitoring implements option 1 in #12. This is a bit inelegant -- it only uses `squeue` because `sacct` wasn't available on the CitC cluster on AWS and uses `pandas` to read in the output from `squeue` which might be overkill. An improvement would be to enable the bot monitoring jobs even after it has crashed (see #23). A future version of the bot could have different implementations for monitoring the jobs and then select the best one based on what is available on a cluster (`sacct` or `squeue`). Also, one might have a look into [PySlurm/pyslurm](https://github.com/PySlurm/pyslurm) to verify if this monitoring can be done more elegantly.

The result of a finished job is checked (existence of slurm output file, existence of tar.gz file, existence of certain pattern in slurm output file). This is relatively clean.

Then the bot adds a comment to the PR with information about the result of the build job. This implements #13.

This PR also addresses #17.

### DIFF
--- a/scripts/eessi-bot-build.slurm
+++ b/scripts/eessi-bot-build.slurm
@@ -127,6 +127,28 @@ else
   echo "eessi-bot-build.slurm: no modules to be loaded"
 fi
 
+source init/minimal_eessi_env  # for $EESSI_PILOT_VERSION and $EESSI_OS_TYPE
+
+software_subdir=
+for i in {1..9};
+do
+  echo "attempt $i to determine software_subdir"
+  raw_output=$(./build_container.sh run ${eessi_tmpdir} python3 /eessi_bot_job/eessi_software_subdir.py ${generic_opt} 2>&1)
+  echo "raw output of eessi_software_subdir.py: '$raw_output'"
+  software_subdir=$(echo $raw_output | tail -1 | sed 's/.* //g')
+  echo "software_subdir: '$software_subdir'"
+  # check if software_subdir has the format:
+  #   CPU_FAMILY/ARCH_IDENTIFIER
+  if [[ $software_subdir == *"/"* ]]; then
+    echo "software_subdir seems well formatted"
+    break
+  else
+    n=$((1<<$i))
+    echo "sleep $n secs (using exponential backoff)"
+    sleep $n
+  fi
+done
+
 # run standard EESSI software-layer install script
 #   in compat layer environment inside build container
 ./build_container.sh run ${eessi_tmpdir} \
@@ -135,12 +157,6 @@ fi
 # create tarball for the above build and place it in some specific
 #   directory on shared disk (just in main directory of job)
 #   see https://github.com/EESSI/eessi-bot-software-layer/issues/7
-
-source init/minimal_eessi_env  # for $EESSI_PILOT_VERSION and $EESSI_OS_TYPE
-
-echo "raw output of eessi_software_subdir.py: '$(./build_container.sh run ${eessi_tmpdir} python3 /eessi_bot_job/eessi_software_subdir.py ${generic_opt} 2>&1)'"
-software_subdir=$(./build_container.sh run ${eessi_tmpdir} python3 /eessi_bot_job/eessi_software_subdir.py ${generic_opt} 2>&1 | tail -1 | sed 's/.* //g')
-echo "software_subdir: '$software_subdir'"
 
 timestamp=$(date +%s)
 
--- a/scripts/eessi-bot-build.slurm
+++ b/scripts/eessi-bot-build.slurm
@@ -44,6 +44,7 @@ while [[ $# -gt 0 ]]; do
   case $1 in
     -g|--generic)
       common_opts="${common_opts} --generic"
+      generic_opt=" --generic"
       shift
       ;;
     -h|--help)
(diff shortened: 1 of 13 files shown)


trz42:
Labelled this as Work-In-Progress to allow for improvements before it gets merged.

Particularly, I thought about the following changes/improvements:

- splitting up current functionality (submit & monitor jobs) into two processes
   - submittter (current bot): submit job in USER_HELD state
   - monitor (new process/tool): pickup USER_HELD jobs by separate process, persist job id in watchlist & release job
- (optional) cronjob that checks whether monitoring process is running
- monitoring job’s main loop:
   - check for user held jobs
   - check jobs in watchlist (use different approach than pandas)
   - handle finished jobs (update PR with job result)
   - handle running jobs (update PR with selected progress/debug info)
- benefits of separate process:
   - simpler bot (processing of events quicker)
   - can also manually submit test jobs

boegel:
@trz42 

i) What's the advantage of submitting a job with a hold set, and letting a separate process release those jobs? Maybe that's asking for trouble... When a job is submitted, we can just let it start straight away, no?

ii) Maybe we should consider that the job script itself triggers a new event for the bot, to signal that a job has completed? That would avoid the need for a bot process that check job statuses in a loop.

iii) The minimal stuff we need to actually start using the bot for building and deploying software in EESSI is the build step (mostly done), and the deploy step (to do). Having some feedback back into the PR is useful, but maybe not strictly necessary? All we would need is that the bot somehow signals that all builds or done (or that one or more have failed), but nothing fancier/more fine-grained than that (we can always check in the bot account itself in some PR ends up in a weird state, for now).

trz42:
i) It can be more safely picked up by the job monitor process. What kind of trouble is this asking for? The main purpose of holding it is to hand it over to the job monitor. If there are better ways, one can do it differently. You can also easily pause/delay any jobs from running by simply killing the job monitor. The other process can still run and process new events.

ii) Hmm, that raises the problem of handling job crashes and communicating with the bot instance (either via the same event channel or some other mechanism). The job script would need more information (about the connection to the bot). What if the bot has crashed in the meantime? The information cannot be sent back to the bot then.

iii) Sure, we don't need very fancy information. I'm currently thinking of something like https://github.com/trz42/software-layer/pull/8 The only thing I need to figure out is how to update some entries in the table. Should not be too hard as PyGithub has a function for updating a comment. My plan is to do smaller improvements to these comments while developing the monitor tool.

trz42:
Getting closer to remove the WIP label. Still need to do some polishing and testing and outline the changes.

scripts/eessi-bot-build.slurm
```diff
@@ -25,27 +20,137 @@
...
+./build_container.sh run ${eessi_tmpdir} \
+    ./install_software_layer.sh $install_software_opts $common_opts "$@"
 
 # create tarball for the above build and place it in some specific
 #   directory on shared disk (just in main directory of job)
 #   see https://github.com/EESSI/eessi-bot-software-layer/issues/7
-EESSI_SILENT=1 source init/eessi_environment_variables
-software_subdir=$(echo ${EESSI_SOFTWARE_SUBDIR} | tr '/' '-')
-timestamp=$(date +%s)
-export TGZ=$(printf "eessi-%s-software-%s-%s-%d.tar.gz" ${EESSI_PILOT_VERSION} ${EESSI_OS_TYPE} ${software_subdir} ${timestamp})
```
boegel, line 48: @trz42 Determining the host CPU arch can now also be done via the build container:

```
cd software-layer
singularity exec -B $PWD:$PWD docker://ghcr.io/eessi/build-node:debian10 python3 $PWD/eessi_software_subdir.py 2>&1 | tail -1
```

or via the `build_container.sh` script:

```
./build_container.sh run /mnt/$USER/EESSI python3 ./eessi_software_subdir.py 2>&1 | tail -1 | sed 's/.* //g'
```

boegel: So, you can revert the changes here to keep using the current `create_tarball.sh` script:

```
source init/minimal_eessi_env  # for $EESSI_PILOT_VERSION and $EESSI_OS_TYPE
software_subdir=$(./build_container.sh run /mnt/$USER/EESSI python3 ./eessi_software_subdir.py 2>&1 | tail -1 | sed 's/.* //g')
```

boegel (changes requested):

scripts/eessi-bot-build.slurm
trz42: Nice!

trz42: Have added `${generic_opt}` when calling the python script (or it would not generate the correct sub directory in case the target is generic).

boegel (changes requested):

boegel (approved):
I've discussed this extensively with @trz42, and tested this thoroughly (see for example https://github.com/boegel/software-layer/pull/7 and https://github.com/boegel/software-layer/pull/8).

There's many things that can be improved, like error handling, code organization, etc., but since this i) works (under normal circumstances), ii) is an *excellent* basis to continue working on, I feel this is ready to go in.

@trz42: Thanks *a lot* for all the effort on this!

Issue #38, opened 2022-10-10
Implement CI tests for bot


trz42:
Add CI tests to make sure changes to the bot don't break it's functionality.

trz42:
CI is good, but it seems very hard to implement an end-to-end test.

PR #55, opened 2022-10-27
add test for mkdir function + CI workflow to run tests with Python 3.6 - 3.11

boegel:
First step towards proper test suite for bot (see #38)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,8 +1,7 @@
 #!/bin/bash
 #
-# This file is part of PyGHee (pronounced as "piggy"), the GitHub Event Executor,
-# is a GitHub App to process GitHub events, implemented in Python;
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# This file is part of PyGHee (pronounced as "piggy"), the GitHub Event Executor,
+# is a GitHub App to process GitHub events, implemented in Python;
+# see https://github.com/boegel/pyghee
```
trz42, line 5: Maybe this should be changed to fit EESSI?

trz42 (changes requested):
Looks ok. Comment in `test.sh` should reflect that this is part of EESSI.

trz42 (approved):

PR #56, opened 2022-10-28
add `run_cmd` method

Hafsa-Naeem:


tasks/build.py
```diff
@@ -176,26 +201,13 @@ def setup_pr_in_arch_job_dir(repo_name, branch_name, pr, arch_job_dir):
                                  shell=True,
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
-    log("Checked out branch!\nStdout %s\nStderr: %s" % (checkout_repo.stdout, checkout_repo.stderr))
-
+    log("Checked out branch!\nStdout %s\nStderr: %s" % (checkout_repo.stdout,checkout_repo.stderr))
```
boegel, line 204: This should become something like:

```python
run_cmd(git_checkout_cmd, "checkout branch '%s'" % branch_name, arch_job_dir)
```

tasks/build.py
```diff
@@ -135,6 +135,38 @@ def create_directory(pr, jobs_base_dir, event_info):
     return ym, pr_id, run_dir
 
 
+def run_cmd(run_command, cmd_detail, arch_job_dir):
```
boegel, line 138: Some suggestions:

```python
def run_cmd(cmd, log_msg='', parent_dir=None):

    if parent_dir is None:
        parent_dir = os.getcwd()

    if log_msg:
      log("'%s' by running '%s' ..." % (log_msg, cmd, ...))
    else:
      log("Running '%s' ..." % (cmd, ...))
```

tasks/build.py
```diff
@@ -135,6 +135,38 @@ def create_directory(pr, jobs_base_dir, event_info):
...
+                                cwd=arch_job_dir,
+                                shell=True,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+    process_output = result.stdout.decode()
+    process_error = result.stderr.decode()
+    process_exit_code = result.returncode
+
+    if process_exit_code != 0:
+        log("Error: '%s' Exitcode: %s" % (process_error,process_exit_code))
```
boegel, line 163: We should always log the exit code, even if it's zero.

Optionally raise an exception if exit code is non-zero:

```python
def run_cmd(cmd, log_msg='', parent_dir=None, raise_error=True):

    log(" '%s' \nexit code: %s\nStdout %s\nStderr: %s" % (cmd_detail, process_exit_code, result.stdout,result.stderr))
    if process_exit_code != 0 and raise_error:
        raise RuntimeError("...." % (cmd, process_exit_code))

**Review by boegel — CHANGES_REQUESTED** [review]
also use `run_cmd` in `submit_job`

File: tasks/build.py (hunk 2 above)
Comment (trz42) [reply tasks/build.py:138]: I wonder if we really need the `log_msg`. Should be enough to simply log the command that is run.

File: tasks/build.py (hunk 3 above)
Comment (trz42) [reply tasks/build.py:163]: While raising exceptions may be good in general, it may complicate the refactoring. Would favour to skip this here. Otherwise we would have to think about when to raise exceptions, when not, etc.

File: tasks/build.py [hunk 4]
```diff
@@ -94,7 +93,7 @@ def get_architecturetargets():
     """get architecturetargets and set arch_target_map
 
     Returns:
-        dict(str, dict): dictionary of arch_target_map which contains entries of the format OS/SUBDIR : ADDITIONAL_SBATCH_PARAMETERS 
+        dict(str, dict): dictionary of arch_target_map which contains entries of the format OS/SUBDIR : ADDITIONAL_SBATCH_PARAMETERS
```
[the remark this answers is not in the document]
boegel:
```
    Returns:
        dict(str, dict): dictionary of arch_target_map which contains entries of the format
                         OS/SUBDIR : ADDITIONAL_SBATCH_PARAMETERS

File: tasks/build.py [hunk 5]
```diff
@@ -135,36 +134,44 @@ def create_directory(pr, jobs_base_dir, event_info):
...
-        log("Error: '%s' Exitcode: %s" % (process_error,process_exit_code))
- 
-    log(" '%s' \nStdout %s\nStderr: %s" % (cmd_detail,result.stdout,result.stderr))
-    
-    return(process_output,process_exit_code)
+        log("Error: '%s' Exitcode: %s" % (process_error, process_exit_code))
+
+    log(" '%s' \nStdout %s\nStderr: %s" % (log_msg, result.stdout, result.stderr))
+
+    return process_output, process_exit_code
```
boegel, line 174: also return `process_error`

tasks/build.py
boegel: OK, let deal with that in issue #6

tests/test_task_build.py
```diff
@@ -32,3 +34,18 @@ def test_mkdir(tmpdir):
...
+
+
+def test_run_cmd(tmpdir):
+    """Tests for run_cmd function."""
+    output, exit_code = run_cmd("echo hello", 'test', tmpdir)
+    assert output == "hello\n"
+    assert exit_code == 0
+
+    output, exit_code = run_cmd("ls -l /does_not_exists.txt", 'fail test', tmpdir)
+    assert exit_code != 0
```
boegel, line 46: `assert "No such file or directory" in error_output`

boegel: Also add a test for a non-existing command:

```python
run_cmd("this_command_does_not_exist", ...)

**Review by boegel — CHANGES_REQUESTED** [review]

File: tasks/build.py [hunk 7]
```diff
@@ -135,6 +135,46 @@ def create_directory(pr, jobs_base_dir, event_info):
...
+    process_output = result.stdout.decode("UTF-8")
+    process_error = result.stderr.decode("UTF-8")
+    process_exit_code = result.returncode
+
+    if process_exit_code != 0:
+        log("Error: '%s' Exitcode: %s" % (process_error, process_exit_code))
+
+    log(" '%s' \nStdout %s\nStderr: %s" % (log_msg, result.stdout, result.stderr))
+
+    return process_output, process_exit_code, process_error
```
boegel, line 175: @Hafsa-Naeem We should make sure the order of return values makes sense, having the exit code return value in the middle is a bit strange I think. So please change to:

```python
    return stdout, stderr, exit_code

```

tasks/build.py
```diff
@@ -135,6 +135,46 @@ def create_directory(pr, jobs_base_dir, event_info):
...
+        log("Running '%s' in directory '%s'" % (cmd, working_dir))
+
+    result = subprocess.run(cmd,
+                            cwd=working_dir,
+                            shell=True,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE)
+    process_output = result.stdout.decode("UTF-8")
+    process_error = result.stderr.decode("UTF-8")
+    process_exit_code = result.returncode
```
boegel, line 168: The `process_` part in the variable names is a bit redundant I think, I would change this to:

```python
    stdout = result.stdout.decode("UTF-8")
    stderr = result.stderr.decode("UTF-8")
    exit_code = result.returncode
```

(and then change below accordingly, of course)

tests/test_task_build.py
```diff
@@ -32,3 +33,19 @@ def test_mkdir(tmpdir):
...
+    assert exit_code == 0
+    assert exit_error == ""
+
+    output, exit_code, exit_error = run_cmd("ls -l /does_not_exists.txt", 'fail test', tmpdir)
+    assert exit_code != 0
+    assert exit_error != ""
+ 
+    output, exit_code, exit_error = run_cmd("this_command_does_not_exist", 'fail test', tmpdir)
+    assert exit_code != 0
+    assert exit_error != ""
```
boegel, line 51: Next to fixing the trivial code style issues reported by the Hound bot (and `flake8`), we should also:
* take into account the changed order of return values for `run_cmd`
* always check exit code + both stdout/stderr in each test case
* have a more specific test and just `!= ""` for the stderr output

Like this:

```python
def test_run_cmd(tmpdir):
    """Tests for run_cmd function."""
    out, err, ec = run_cmd("echo hello", 'test', tmpdir)
    assert ec == 0
    assert out == "hello\n"
    assert err == ""

    out, err, ec = run_cmd("ls -l /does_not_exists.txt", 'fail test', tmpdir)
    assert ec != 0
    assert out == ""
    assert "No such file or directory" in err

    out, err, ec = run_cmd("this_command_does_not_exist", 'fail test', tmpdir)
    assert ec != 0
    assert out == ""
    assert "this_command_does_not_exist: command not found" in err

boegel (changes requested):

boegel (approved):
lgtm

PR #59, opened 2022-11-09
move run_cmd function from tasks/build.py to 'tools' module

trz42:
Small PR which creates a new file `tools/cmds.py` to which the function `run_cmd` from `tasks/build.py` is moved. This should make it easier to reuse `run_cmd` in other components of the bot.

The function was also tweaked a bit (documentation, etc.).


tools/cmds.py
```diff
@@ -0,0 +1,57 @@
...
+                            stderr=subprocess.PIPE)
+    stdout = result.stdout.decode("UTF-8")
+    stderr = result.stderr.decode("UTF-8")
+    exit_code = result.returncode
+
+    if exit_code != 0:
+        log(f"run_cmd(): Error running '{cmd}' in '{working_dir}\n"
+            f"           stdout '{stdout}'\n"
+            f"           stderr '{stderr}'\n"
+            f"           exit code {exit_code}")
```
boegel, line 55: @trz42 Now we're no longer logging the output if the command was successful, we should keep doing that?

boegel (changes requested):
I also wonder if we can come up with a better name than `tools.cmds`...

Maybe something more generic? `cmds` is very specific (what other functions would we have in there?)...

The best I can come up with is `tools/utils` though, which is way too generic (and perhaps confusing too)...

Maybe we can park `run_cmd` in `tools/__init__.py`, so we can do `from tools import run_cmd`?

tools/cmds.py
trz42: Hmm. We can do that. As it was, it was logging twice in case of error.

trz42:
Yeah, `utils` sounds too generic (could be anything). Was thinking about `shell` or `exec`. Parking in `__init__.py` feels not so good to me.

tools/cmds.py
trz42: Also logging in case of "success" (exit code equals zero). It also logs `stderr` which might be counterintuitive at first, but who knows if a command is not printing something on `stderr`.

boegel: Makes sense to always log `stderr` (in case of warnings, etc.)

boegel (approved):
lgtm

PR #60, opened 2022-11-09
make bot ignore non-bot jobs

Hafsa-Naeem:


boegel:
@Hafsa-Naeem We should move the code block that creates the symlink and the code block that releases the job *after* checking whether the metadata file exists in `process_new_job`

eessi_bot_job_manager.py
```diff
@@ -27,6 +27,7 @@
 import re
 import subprocess
 import time
+import os.path
```
trz42, line 30: Just curious, do we need to import `os.path`? `os` is imported already (a few lines above).

boegel: Then no, just `import os` is sufficient to use `os.path`

boegel:
closing in favor of #63

Issue #64, opened 2022-11-14
get flake8 to pass on all .py files


boegel:
Ideally using a separate PR per Python file

* [x] `connections/github.py`
* [x] `eessi_bot_event_handler.py`
* [x] `eessi_bot_job_manager.py` (see #63)
* [x] `tasks/build.py` (done?)
* [x] `tests/test_eessi_bot_job_manager.py`
* [x] ` tests/test_task_build.py`
* [x] `tools/args.py`
* [x] `tools/config.py`
* [x] `tools/logging.py`

We should also add a code style check in GitHub Actions, cfr. https://github.com/easybuilders/easybuild-framework/blob/develop/.github/workflows/linting.yml

trz42:
All tasks done, all PRs merged. Issue can be closed.

PR #69, opened 2022-11-14
Pass flake8 tests for tests/test_task_build.py

bedroge:
See https://github.com/EESSI/eessi-bot-software-layer/issues/64.


trz42 (approved):
Looks good. This seems to be the only issue `flake8` reported for this file. See tests in #70.

PR #70, opened 2022-11-14
Add flake8 checks

trz42:
Addresses part of #64

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,11 +16,17 @@ jobs:
         with:
           python-version: ${{matrix.python}}
 
-      - name: Install required Python packages + pytest
+      - name: Install required Python packages + pytest + flake8
         run: |
+          python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -16,11 +16,17 @@ jobs:
...
           python -m pip install pytest
+          python -m pip install --upgrade flake8
 
       - name: Run test suite
         run: |
           ./test.sh
+
+      - name: Run flake8 to verify PEP8-compliance of Python code
+        run: |
+          flake8
```
boegel, line 32: It may make sense to use a separate workflow for the tests and flake8, so we always get the result of both in each PR.

Currently, if the tests fail, flake8 will not be run. So if you manage to get the tests to pass, you still won't be done since then the code style dance begins...

We can tackle that in a follow-up PR, I'll merge this as is for now, since it already helps.

boegel (approved):

PR #72, opened 2022-11-16
Pass flake8 tests for tasks/build.py

trz42:
Three simple changes only.


trz42:
Was started from wrong branch. Replaced by #74

PR #77, opened 2022-11-17
Error handling

jacobz137:
The following changes have been made:
 - On error, run_cmd raises a RuntimeError in addition to logging the error message.
 - Slurm errors are logged in the case of states "Failure", "Out of Memory" or "Time Out". I opted not to log issues due to "Pending: ReqNodeNotAvail", because this can either mean that the node is fully in use and is unable to run any more jobs, which isn't an error, or it can mean that there is a problem with the node and it is offline. I did not see a way to differentiate between these two cases. A "warning" could be appropriate here, but I thought that may be annoying if it happens too often.
 - Errors are logged when token creation fails, due to missing configuration parameters.
 - From what I could tell, /tools/config.py was using the wrong method to read the config file, but maybe I misunderstood the purpose.
 - The mkdirs method defined in /tasks/build.py and again in eessi_bot_job_manager.py looks like it can be accomplished with os.makedirs' exist_ok parameter.

If there's anything else that should go into this, let me know.

### DIFF
--- a/connections/github.py
+++ b/connections/github.py
@@ -11,7 +11,7 @@
 #
 import datetime
 
-from tools import config
+from tools import config, logging
 from github import Github, GithubIntegration
 
 _token = None
@@ -28,9 +28,15 @@ def get_token():
     with open(private_key_path, 'r') as private_key_file:
         private_key = private_key_file.read()
 
-    github_integration = GithubIntegration(app_id, private_key)
-    # Note that installation access tokens last only for 1 hour, you will need to regenerate them after they expire.
-    _token = github_integration.get_access_token(installation_id)
+    # If the config keys are not set, get_access_token will raise a NotImplementedError
+    # Returning NoneType token will stop the connection in get_instance
+    try:
+        github_integration = GithubIntegration(app_id, private_key)
+        # Note that installation access tokens last only for 1 hour, you will need to regenerate them after they expire.
+        _token = github_integration.get_access_token(installation_id)
+    except NotImplementedError as e:
+        logging.error(e)
+        _token = None
 
     return _token
 
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -84,7 +84,7 @@ class EESSIBotSoftwareLayerJobManager:
                     "reason": job[8],
                 }
                 if current_jobs[job[0]]['state'] in bad_state_codes:
-                    error("Job {} in state {}: {}".format(job[0], job[4], bad_state_messages[job[4]))
+                    error("Job {} in state {}: {}".format(job[0], job[4], bad_state_messages[job[4]]))
 
         return current_jobs
 
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -89,7 +89,7 @@ class EESSIBotSoftwareLayerJobManager:
                     "reason": job[8],
                 }
                 if state in bad_state_messages:
-                    error("Job {} in state {}: {}".format(job_id, state, bad_state_messages[state]))
+                    log("Job {} in state {}: {}".format(job_id, state, bad_state_messages[state]))
 
         return current_jobs
 
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -72,6 +72,8 @@ class EESSIBotSoftwareLayerJobManager:
         lines = str(squeue.stdout, "UTF-8").rstrip().split("\n")
         bad_state_messages = {'F': 'Failure', 'OOM': 'Out of Memory', 'TO': 'Time Out'}
         bad_state_codes = bad_state_messages.keys()
+
+        # get job info, logging any Slurm issues
         for i in range(2, len(lines)):
             # assume lines 2 to len(lines) contain jobs
             job = lines[i].rstrip().split()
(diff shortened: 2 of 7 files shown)


connections/github.py
```diff
@@ -28,9 +28,15 @@ def get_token():
...
-    # Note that installation access tokens last only for 1 hour, you will need to regenerate them after they expire.
-    _token = github_integration.get_access_token(installation_id)
+    # If the config keys are not set, get_access_token will raise a NotImplementedError
+    # Returning NoneType token will stop the connection in get_instance
+    try:
+        github_integration = GithubIntegration(app_id, private_key)
+        # Note that installation access tokens last only for 1 hour, you will need to regenerate them after they expire.
+        _token = github_integration.get_access_token(installation_id)
+    except NotImplementedError as e:
+        logging.error(e)
```
boegel, line 38: Please don't use single-letter variables:

```suggestion
    except NotImplementedError as err:
        logging.error(err)

File: eessi_bot_job_manager.py [hunk 2]
```diff
@@ -73,9 +68,12 @@ def get_current_jobs(self):
         # if any with the following information per job:
         #  jobid, state, nodelist_reason
         # skip first two lines of output ("range(2,...)")
-        # TODO check for errors of squeue call
         current_jobs = {}
         lines = str(squeue.stdout, "UTF-8").rstrip().split("\n")
+        bad_state_messages = {'F': 'Failure', 'OOM': 'Out of Memory', 'TO': 'Time Out'}
```
boegel, line 73:
```suggestion
        bad_state_messages = {
            "F": "Failure",
            "OOM": "Out of Memory",
            "TO": "Time Out",
        }

File: eessi_bot_job_manager.py [hunk 3]
```diff
@@ -85,6 +83,8 @@ def get_current_jobs(self):
                     "state": job[4],
                     "reason": job[8],
                 }
+                if current_jobs[job[0]]['state'] in bad_state_codes:
+                    error("Job {} in state {}: {}".format(job[0], job[4], bad_state_messages[job[4]]))
```
boegel, line 87: I'm not sure if we should make the job manager *crash* if it hits a job that has timed out, for example...
Instead, the job manager should report that back to the PR, and continue, ideally?

Code-style wise: we're accessing the "`state`" value in multiple ways here, that could be done better.
Also, we should report an error in the `else` case, i.e. if don't get exactly 9 "fields"?

Something like:

```python
if len(job) == 9:
    job_id, job_state, job_state_reason = job[0], job[4], job[8]
    current_jobs[job_id] = {
        "jobid": job_id,
        "state": job_state,
        "reason": job_reason,
    }
    if job_state in bad_state_codes:
        error("Job {} in state {}: {}".format(job_id, job_state, bad_state_messages[job_state]))
else:
    error("Incorrect job info, expected 9 fields, found {}: {}".format(len(job), job))

File: tools/__init__.py [hunk 4]
```diff
@@ -18,6 +18,41 @@
...
+    Returns:
+        tuple of 3 elements containing
+        - stdout (string): stdout of the process
+        - stderr (string): stderr of the process
+        - exit_code (string): exit code of the process
+    """
+
+    stdout, stderr, exit_code = run_subprocess(cmd, log_msg, working_dir)
+
+    if exit_code != 0:
```
boegel, line 37: Rather than breaking up the existing `run_cmd` function into two functions, and causing confusion on when to use which, how about adding a `raise_on_error` named argument to `run_cmd`, and letting that determine whether an error is only logged or it triggering the raising of a `RuntimeError`?

boegel (changes requested):

tasks/build.py
```diff
@@ -136,13 +126,13 @@ def create_pr_dir(pr, jobs_base_dir, event_info):
     pr_id = 'pr_%s' % pr.number
     event_id = 'event_%s' % event_info['id']
     event_dir = os.path.join(jobs_base_dir, ym, pr_id, event_id)
-    mkdir(event_dir)
+    os.makedirs(event_dir, exist_ok=True)
```
boegel, line 129: Using `os.makedirs` makes total sense. This wasn't supported yet in Python 2.7, so I guess force of habit made me go with the usual `mkdir` approach :)

tools/config.py
```diff
@@ -20,9 +20,9 @@ def read_file(path):
     Read a given configuration file.
     """
     global _config
-    _config = configparser.ConfigParser()
     try:
-        _config.read(path)
+        _config = configparser.ConfigParser()
+        _config.read_file(path)
```
boegel, line 25: `read` works just as well, because (from https://docs.python.org/3.6/library/configparser.html#configparser.ConfigParser.read):
```
If filenames is a string or [path-like object](https://docs.python.org/3.6/glossary.html#term-path-like-object), it is treated as a single filename.
```

but `read_file` is clearer, so :+1:

eessi_bot_job_manager.py
jacobz137: Is there any possibility that there won't be 9 fields? I thought Slurm output always had the same fields.

trz42: Hmm, the number of output fields could be depending on the configuration or build of Slurm.

tools/config.py
trz42: Actually, `read_file` crashes the event handler when reading `app.cfg`.

trz42:
Critical change requests have been implemented with commits up to https://github.com/EESSI/eessi-bot-software-layer/pull/77/commits/548ff5312c4bd27aef2e84f49cc90e872e0e4589

Conflicts due to more recent upstream changes are handled by wrapper PR #129

PR #78, opened 2022-11-17
implemented the run_cmd everywhere

### LINKED ISSUE #61: use run_cmd everywhere (instead of using subprocess.run)
After #59 is merged

For example in `eessi_bot_job_manager.py` (maybe also in other places)

Hafsa-Naeem:
fixes #61

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -31,12 +31,11 @@ import configparser
 import glob
 import os
 import re
-import subprocess
 import time
 
 from connections import github
 from datetime import datetime, timezone
-from tools import args, config
+from tools import args, config, run_cmd
 
 from pyghee.utils import log, error
 
@@ -54,28 +53,18 @@ class EESSIBotSoftwareLayerJobManager:
         username = os.getlogin()
 
         squeue_cmd = "%s --long --user=%s" % (self.poll_command, username)
-        log(
-            "get_current_jobs(): run squeue command: %s" % squeue_cmd,
-            self.logfile,
+        squeue_output, squeue_err, squeue_exitcode = run_cmd(
+            squeue_cmd,
+            "get_current_jobs(): squeue command"
         )
 
-        squeue = subprocess.run(
-            squeue_cmd,
-            shell=True,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        )
-        log(
-            "get_current_jobs(): squeue output\n%s" % squeue.stdout,
-            self.logfile,
-        )
         # create dictionary of jobs
         # if any with the following information per job:
         #  jobid, state, nodelist_reason
         # skip first two lines of output ("range(2,...)")
         # TODO check for errors of squeue call
         current_jobs = {}
-        lines = str(squeue.stdout, "UTF-8").rstrip().split("\n")
+        lines = str(squeue_output).rstrip().split("\n")
         for i in range(2, len(lines)):
             # assume lines 2 to len(lines) contain jobs
             job = lines[i].rstrip().split()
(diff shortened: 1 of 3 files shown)


trz42 (commented):
This looks like a solid PR that could be merged quickly. Given recent experience with other PRs, it would be good to test the PR e.g., `squeue` and `scontrol` still working as before

I can give a try some time on Sunday or Tuesday.

I looked a bit into the `subprocess` module. It seems `subprocess.run()` returns a `subprocess.CompletedProcess` object which includes `stdout`, `stderr` and `returncode` (see https://docs.python.org/3.11/library/subprocess.html?highlight=subprocess#subprocess.CompletedProcess). I think we good make good use of that in a subsequent PR.

Moreover, `subprocess.run` has a parameter for the encoding used for `stdout` and `stderr` which we could use to avoid the decoding step (`"UTF-8"`).

trz42:
Oh, interesting find. Since `run_cmd` does not have a parameter for the log file, logging from `run_cmd` in the job manager now actually appears in the event handler's log. This wasn't a problem when `run_cmd` was used by the event handler.

A simple / quick fix would probably be, the `run_cmd` needs to have another parameter `logfile=None`. If a value is provided, calls to `log` in `run_cmd` need to add that. In the long term one could have a `log` function/method in the classes for the event handler and the job manager, which then would add the log file name.

Also, some other logging is not right (written into wrong log file, not using 'f' for f-string, mispelled variable name). These issues are fixed in two commits included in #79:
- https://github.com/EESSI/eessi-bot-software-layer/pull/79/commits/a288a643664813560376b2128e06eb4e488d2c09
- https://github.com/EESSI/eessi-bot-software-layer/pull/79/commits/5f2860a7382798e2b052210f5ff615079e69e34c

In addition, a test may need to be fixed with

https://github.com/EESSI/eessi-bot-software-layer/pull/79/commits/5d887017cd9cfcb4140093bd8de2c6d06ec97f07

trz42:
Does this PR close any issues? Good to add them with 'closes #issue_number'

eessi_bot_job_manager.py
```diff
@@ -54,28 +53,18 @@ def get_current_jobs(self):
         username = os.getlogin()
 
         squeue_cmd = "%s --long --user=%s" % (self.poll_command, username)
-        log(
-            "get_current_jobs(): run squeue command: %s" % squeue_cmd,
-            self.logfile,
```
boegel, line 59: @Hafsa-Naeem As @trz42 pointed out, we should enhance `run_cmd` so a `logfile` can be passed into it, which is used in the `log` statements there:

```python
run_cmd(squeue_cmd, ..., logfile=self.logfile)

**Review by boegel — CHANGES_REQUESTED** [review]

File: eessi_bot_job_manager.py (hunk 1 above)
Comment (boegel) [reply eessi_bot_job_manager.py:59]: Also add a test if you enhance `run_cmd`:

```python
log_file = os.path.join(tmpdir, "log.txt")
run_cmd("echo hello", "test in file", tmpdir, log_file=log_file)
with open(log_file, "r") as fp:
    assert "test in file" in fp.read()

File: eessi_bot_job_manager.py [hunk 2]
```diff
@@ -54,28 +53,18 @@ def get_current_jobs(self):
...
-        )
-
-        squeue = subprocess.run(
+        squeue_output, squeue_err, squeue_exitcode = run_cmd(
             squeue_cmd,
```
boegel, line 57: @Hafsa-Naeem To make sure that logging is done in the right log file, we need to pass down `self.logfile` here:

```python
        squeue_output, squeue_err, squeue_exitcode = run_cmd(
            squeue_cmd,
            "get_current_jobs(): squeue command",
            log_file=self.logfile,
        )
```

Same for the other `run_cmd` calls in `eessi_bot_job_manager.py`

boegel (changes requested):

boegel (approved):

PR #80, opened 2022-11-23
split up option parsers for event handler and job manager

### LINKED ISSUE #25: Separate handling of command line arguments
`tools/args.py` defines the command line arguments. Currently this is used for two programs where not all arguments are used. Would be good to support different sets of arguments for different programs/tools. In the future, we may add more tools which have other command line arguments.

Hafsa-Naeem:
Closes #25

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -37,7 +37,7 @@ import time
 from connections import github
 from tools.args import parse_common_args, job_manager_parse
 from datetime import datetime, timezone
-from tools import args, config
+from tools import config
 
 from pyghee.utils import log, error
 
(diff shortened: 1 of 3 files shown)


eessi_bot_job_manager.py
```diff
@@ -532,7 +538,8 @@ def process_finished_job(self, finished_job):
 
 def main():
     """Main function."""
-    opts = args.parse()
+    opt = EESSIBotSoftwareLayerJobManager()
+    opts = opt.parse_args()
```
boegel, line 542: If `job_manager_parse` calls `parse_common_args`, this becomes a lot simpler:

```python
opts = job_manager_parse()

boegel:
Correct split is:

* common:
  * --debug (new option)
* event handler specific:
  * --cron
  * --port
  * --file
  * --build
  * --test
* job-manager specific
  * --jobs
  * --max-manager-iterations

trz42:
Looks good.

PR #82, opened 2022-11-28
Update pr comments

### LINKED ISSUE #32: Function for identifying PR comment to be updated
The bot reports status changes, results, etc as comments back to the PR. For updating these comments it first needs to identify the comment. Currently (as of PR #24) this is based on the contents of the first line in a PR comment (using a job's ID and the app's name which is defined in `app.cfg`). The code is duplicated at some places and should be collected into one function, e.g.,

`def identify_pull_request_comment(pattern: str, repository: str, pr_number: int) -> Tuple[bool, int]:`

which receives a search pattern, a repository name and a pull request number as input and returns a tuple with a return value (`True` - comment found, `False` - no matching comment found) and a number identifying the first comment to be updated (only meaningful if return value is `True`).

Hafsa-Naeem:
Closes #32

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -16,7 +16,7 @@
 import waitress
 
 from connections import github
-from tools import args, config
+from tools import config
 from tools.args import parse_common_args, event_handler_parse
 from tasks.build import submit_build_jobs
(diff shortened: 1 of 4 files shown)


trz42:
This PR is replaced by #83

So closing this one.

PR #83, opened 2022-11-30
identify pr comment to be updated

### LINKED ISSUE #32: Function for identifying PR comment to be updated

Hafsa-Naeem:
fixes #32

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -33,9 +33,11 @@ import os
 import re
 import time
 
+
 from connections import github
 from datetime import datetime, timezone
 from tools import args, config, run_cmd
+from tools.pr_comments import get_comment, update_comment
 
 from pyghee.utils import log, error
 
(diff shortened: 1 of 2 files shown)


tools/pr_comments.py
```diff
@@ -0,0 +1,35 @@
...
+# author: Bob Droege (@bedroge)
+# author: Hafsa Naeem (@hafsa-naeem)
+# author: Thomas Roeblitz (@trz42)
+#
+# license: GPLv2
+#
+import re
+
+
+def get_comment(pr, job_id):
```
trz42, line 17: I'd suggest to replace `job_id` with `search_pattern` as a parameter to the function. Also see comment about lines 23 & 24 below.

Also docstring for the function would be good.

tools/pr_comments.py
```diff
@@ -0,0 +1,35 @@
...
+    comments = pr.get_issue_comments()
+    for comment in comments:
+        # NOTE adjust search string if format changed by event
+        #        handler (separate process running
+        #        eessi_bot_event_handler.py)
+        cms = f".*submitted.*job id `{job_id}`.*"
+        # cms = f".*{search_pattern}.*"
```
trz42, line 24: I'd suggest to use a generic version, for example, the one that is commented out in line 24 now. If we need a more specific one, we could always define an additional function, say `get_submitted_job_comment()` which could be defined as follows

```python
def get_submitted_job_comment(pr, job_id):
    search_pattern = f"submitted.*job_id `{job_id}`"
    return get_comment(pr, search_pattern)
```

tools/pr_comments.py
```diff
@@ -0,0 +1,35 @@
...
+
+        comment_match = re.search(cms, comment.body)
+
+        if comment_match:
+            return comment
+    return None
+
+
+def update_comment(cmnt_id, pr, update):
```
trz42, line 33: Docstring for the function would be good to have.

eessi_bot_job_manager.py
```diff
@@ -339,23 +330,16 @@ def process_finished_job(self, finished_job):
...
         # determine comment to be updated
         if "comment_id" not in finished_job:
-            comments = pull_request.get_issue_comments()
-            for comment in comments:
-                # NOTE adjust search string if format changed by event
-                #        handler (separate process running
-                #        eessi_bot_event_handler.py)
-                cms = f".*submitted.*job id `{finished_job['jobid']}`.*"
+            # search_pattern = f"submitted.*job id `{finished_job['jobid']}`"
+            finished_job_cmnt = get_comment(pull_request, finished_job['jobid'])
```
trz42, line 334: If parameters to function `get_comment` are changed to `get_comment(pr, search_pattern)` the call here needs to be changed to.

eessi_bot_job_manager.py
```diff
@@ -250,34 +252,23 @@ def process_new_job(self, new_job):
...
-                    if comment_match:
-                        log(
-                            "process_new_job(): found comment with id %s"
-                            % comment.id,
-                            self.logfile,
-                        )
-                        new_job["comment_id"] = comment.id
-                        break
+                # search_pattern = f"submitted.*job id `{new_job['jobid']}`"
+                new_job_cmnt = get_comment(pr, new_job['jobid'])
```
trz42, line 256: If parameters to function `get_comment` are changed to `get_comment(pr, search_pattern)` the call here needs to be changed to.

trz42 (commented):
Good progress. To make the function `get_comment` also reusable for other comments it would be good to use a search pattern as parameter instead of a job id. If a function using a job id is needed, that could then call the more generic one.

trz42:
It's almost done. A few small things to adjust (typos, variable types, completing docstrings). If also the conflicts are resolved, the PR can be merged.

trz42 (commented):
More comments. Sorry for delay in submitting the review.

trz42:
Thanks for the changes addressing comments. PR is ready to get merged.

PR #84, opened 2022-11-30
determine_username

### LINKED ISSUE #81: Change method for determining name of user that runs the job manager
The job manager uses
```python
username = os.getlogin()
```

See job manager code below where the above is used
https://github.com/EESSI/eessi-bot-software-layer/blob/05716de42efdf5d9c4f3bc1dc8a593ba0fea0160/eessi_bot_job_manager.py#L56

In some cases, for example, when `sudo`ing into a user account (say `newuser`), `os.getlogin()` does not return the expected value. `os.getlogin()` should be replaced by a different method.

Hafsa-Naeem:
fixes #81

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -53,7 +53,9 @@ class EESSIBotSoftwareLayerJobManager:
 
     def get_current_jobs(self):
         # who am i
-        username = os.getlogin()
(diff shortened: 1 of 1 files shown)


trz42:
Looks good to me. Tested via `sudo -i`, then `root` is returned (old method returned the account which ran `sudo`). No changes needed.

Issue #39, opened 2022-10-10
Remove dependency on `curl`


trz42:
The bot uses the command `curl` to fetch a patch file from GitHub. Should we remove that dependency?

boegel:
Running the `curl` command can probably be replaced by using `requests.get`, see also https://365datascience.com/tutorials/python-tutorials/python-requests-package/

PR #85, opened 2022-12-02
tool to resubmit a build job locally

### LINKED ISSUE #29: Add a tool for resubmitting jobs locally
If a job fails, it may be convenient to investigate only that single job (on a single bot instance). This could be done by updating the PR and reconfiguring and restarting the bot (or all bots at possibly different sites) such that when the `bot:label` event is being resent only that particular job is rerun. This is a bit cumbersome.

It would be nicer if one could resubmit the job locally (with or without some local modifications without changing the PR). To document progress, the reprocessing could still add comments to the original job comment of the PR (adding more rows in the status table).

trz42:
Initial version reimplementing the tool based on recent cleanups and enhancements to the bot sources.

- [x] test if it works
  - works, see for example https://github.com/trz42/software-layer/pull/49#issuecomment-1325473297 (the build job still fail, but the resubmission works as intended)
- [x] add documentation about how to use it
- [ ] address review
- [ ] update code with recently merge pull requests (not all below may be relevant, just check nevertheless)
  - [ ] #130
  - [ ] #129
  - [ ] #124
  - [ ] #123
  - [ ] #116 
  - [ ] #84
  - [ ] #83
  - [ ] #80
  - [ ] #78  

Closes #29

Uses a different method to obtain a patch file. Using that method would help solving issue #39


tools/config.py
```diff
@@ -28,6 +65,69 @@ def read_file(path):
...
+        error(f"{fn}(): Unable to read configuration file {path}!")
+
+    # Check if all required configuration sections and options can be found.
+    # Also, process some values that are not just strings.
+    for section in required_config.keys():
+        if section not in config:
+            error(f"{fn}(): Missing section '{section}' in config file {path}.")
+        for item in required_config[section]:
+            if item not in config[section]:
+                error(f"{fn}(): Missing item '{item}' in section '[{section}]' "
```
boegel, line 95: Rather than making this a hard error, you could also just pass through to a default value if the configuration setting is not specified?

```python
value = config[section].get(item, None)
if value is None:
    ...

Comment (trz42) [reply tools/config.py:95]: Possibly. Wonder if we can always specify a default value. But we can try.

If the configuration is only read at the beginning, crashing might be ok.

**Comment by trz42** [comment]
Seems the resubmit script adds http proxy command line arguments to the sbatch command. See output

```
run_cmd(): Running '/cm/shared/apps/slurm/20.02.7/bin/sbatch --hold --time=1-00:00:00 --nodes=1 --ntasks-per-node=8 --partition=defq /home/thomarob/pilot.nessi/PR62/scripts/eessi-bot-build.slurm --tmpdir /work/$USER/\$SLURM_JOB_ID/EESSI --http-proxy  --https-proxy  --load-modules singularity-ce/3.10.0' in directory '/home/thomarob/pilot.nessi/PR62/jobs/2022.11/pr_49/408107/rerun_005'
```

This is then propagated to the installation script. Some added debug information printed by `EESSI-pilot-install-software.sh` is

```
command line arguments: '--http-proxy --https-proxy'
http_* env vars at start of script
http_* env vars after parsing command line arguments
http_proxy=--https-proxy
```

This causes problems for EasyBuild to fetch (some) sources with the (error) message

```
== 2022-12-21 00:03:25,253 filetools.py:840 WARNING IOError occurred while trying to download http://download.oracle.com/berkeley-db/db-18.1.40.tar.gz to /tmp/thomarob/easybuild/sources/d/DB/db-18.1.40.tar.gz: <urlopen error [Errno -2] Name or service not known>
``` 

The `http*` parameters are added through https://github.com/trz42/eessi-bot-software-layer/blob/feature/new-resubmit-tool/resubmit.py#L403..L406

While the configuration file has the http proxy settings commented out, e.g.,

```ini
#http_proxy = http://PROXY_DNS:3128/
#https_proxy = http://PROXY_DNS:3128/
```

We need to figure out where and how the code in `resubmit.py` need to be changed (just when using the config settings https://github.com/trz42/eessi-bot-software-layer/blob/feature/new-resubmit-tool/resubmit.py#L403..L406, or also when reading the settings https://github.com/trz42/eessi-bot-software-layer/blob/feature/new-resubmit-tool/resubmit.py#L468).

trz42:
It seems changing the lines

```python
    if buildenv[config.OPTION_HTTP_PROXY] is not None:
        command_line += f" --http-proxy {buildenv[config.OPTION_HTTP_PROXY]}"
    if buildenv[config.OPTION_HTTPS_PROXY] is not None:
        command_line += f" --https-proxy {buildenv[config.OPTION_HTTPS_PROXY]}"
    if buildenv[config.OPTION_LOAD_MODULES] is not None:
        command_line += f" --load-modules {buildenv[config.OPTION_LOAD_MODULES]}"
```

to

```python
    if len(buildenv[config.OPTION_HTTP_PROXY]) > 0:
        command_line += f" --http-proxy {buildenv[config.OPTION_HTTP_PROXY]}"
    if len(buildenv[config.OPTION_HTTPS_PROXY]) > 0:
        command_line += f" --https-proxy {buildenv[config.OPTION_HTTPS_PROXY]}"
    if len(buildenv[config.OPTION_LOAD_MODULES]) > 0:
        command_line += f" --load-modules {buildenv[config.OPTION_LOAD_MODULES]}"
```

Seems that this works, see last resubmitted job (id `421943 `) in https://github.com/trz42/software-layer/pull/49#issuecomment-1325473297

trz42:
This is quite old and outdated. If we need such functionality later, we can revisit this PR.

PR #117, opened 2022-12-08
Add CI security action

ocaisa:


.github/workflows/tests.yaml
```diff
@@ -1,18 +1,20 @@
 name: Run tests
 on: [push, pull_request]
+# Declare default permissions as read only.
+permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9, '3.10', '3.11']
```
boegel, line 8: @ocaisa It's a bit of a shame we have to drop testing with Python 3.6, since that's the default system Python in RHEL8 and derivatives...

Were the tests failing with Python 3.6?

ocaisa: Yeah the problem is using ubuntu-latest, you need ubuntu-20.04 to get python 3.6 (IIRC)

ocaisa: So no, tests not failing, just python action can't give you python version

boegel: fixed in #122

PR #123, opened 2022-12-12
improve start of the app

### LINKED ISSUE #9: Improve start of app (in PyGHee)

Hafsa-Naeem:
fixes #9

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -23,6 +23,7 @@ from tasks.deploy import deploy_built_artefacts
 
 from pyghee.lib import PyGHee, create_app, get_event_info, read_event_from_json
 from pyghee.utils import log
+import os
 
 
 class EESSIBotSoftwareLayer(PyGHee):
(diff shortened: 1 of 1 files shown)


eessi_bot_event_handler.py
```diff
@@ -91,6 +92,19 @@ def handle_pull_request_event(self, event_info, log_file=None):
         else:
             log("No handler for PR action '%s'" % action)
 
+    def start(self, app, port=None):
```
boegel, line 95: Please use `3000` instead of `None` as default

```python
    def start(self, app, port=3000):

**Review by boegel — CHANGES_REQUESTED** [review]

File: eessi_bot_event_handler.py (hunk 1 above)
Comment (boegel) [reply eessi_bot_event_handler.py:95]: And also add a docstring to the `start` method here

File: eessi_bot_event_handler.py [hunk 2]
```diff
@@ -91,6 +92,19 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+        start_msg = "EESSI bot for software layer started!"
+        print(start_msg)
+        log(start_msg)
+        port_info = "app is listening on port %s" % port
+        print(port_info)
+        log(port_info)
+        my_logfile = os.path.join(os.getcwd(), "pyghee.log")
```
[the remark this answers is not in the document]
boegel: see issue #125

PR #124, opened 2022-12-12
check build permission

### LINKED ISSUE #97: Limit which GH accounts can trigger build jobs
Similarly to the deploy step, the build step could be restricted to a configurable list of GH accounts. This would provide an additional protection against running any code on bot instances.

Hafsa-Naeem:
fixes #97

### DIFF
--- a/app.cfg.example
+++ b/app.cfg.example
@@ -100,6 +100,9 @@ upload_policy = once
 # value can be a space delimited list of GH accounts
 deploy_permission = trz42
 
+[buildcfg]
+# which GH account has the permission to trigger the build 
+build_permission = Hafsa-Naeem
 
 [architecturetargets]
 # defines both for which architectures the bot will build
--- a/app.cfg.example
+++ b/app.cfg.example
@@ -67,6 +67,11 @@ slurm_params = --hold
 # full path to the job submission command
 submit_command = /usr/bin/sbatch
 
+# which GH account has the permission to trigger the build (by setting
+# the label 'bot:build' (apparently this cannot be restricted on GitHub)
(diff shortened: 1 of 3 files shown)


tasks/build.py
```diff
@@ -389,3 +392,25 @@ def submit_build_jobs(pr, event_info):
         create_metadata(job, repo_name, pr, job_id)
         # report submitted jobs (incl architecture, ...)
         create_pr_comments(job, job_id, app_name, job_comment, pr, repo_name, gh, symlink)
+
+
+def build_built_artefacts(pr, event_info):
```
boegel, line 397: Rework this as `check_build_permission` which just returns `False` or `True`, and does logging, but does not call `submit_build_jobs`

And then in `handle_pull_request_labeled_event`:

```python
if check_build_permission(pr, event_info):
    submit_build_jobs(pr, event_info)
```

And make sure to add a docstring for `check_build_permission` here

app.cfg.example
```diff
@@ -100,6 +100,9 @@ upload_policy = once
 # value can be a space delimited list of GH accounts
 deploy_permission = trz42
 
+[buildcfg]
+# which GH account has the permission to trigger the build 
+build_permission = Hafsa-Naeem
```
boegel, line 105: Let's add `build_permission` to the existing `buildenv` section, rather than introducing a new `buildcfg` section

boegel (changes requested):
Please also add a short paragraph in the README on the `build_permission` setting (in the `buildenv` section of the README)

boegel (approved):

Issue #96, opened 2022-12-03
Control center in main PR comment


trz42:
Provide an overview of the status of the PR and means to control instances (disable/enable instances and/or targets, deploy builds for specific instance/target).

Could be implemented by a "master" bot instance which receives all events (PR opened, labelled and particularly updates to comments).

Add mockups of the layout in subsequent comments to this issue.

trz42:
**Bot instance at `name` with defined targets:**
|arch|repo/version|status|build progress|deploy progress|
|---|---|---|--------|--------|
|`x86_64/generic`|`NESSI/2022.11`| <ul><li>- [ ] enabled</li></ul> | ✅ | :red_cross: |
<hr/>

**Last state changes ($n$ per target)**
<details>
<summary>All state changes (human readable, table)</summary>

|date|arch+repo/version|job id|action/state|details|
|----|--------|-----|-----|------------------------|
|today|`aarch64+NESSI/2022.11`|`1234`|`submitted`|job dir `/tmp/foo`|
</details>
<details>
<summary>All state changes (machine readable, `json`)</summary>

```json
{ "date 1" }
```
</details>
<hr/>

**control room**
_Note, add commands to the bot below. Keep the ones which are there. Add new at the end of the list. See DOC for a list of supported commands._
`Example: @bot cancel job 12345`

bedroge:
While reviewing https://github.com/EESSI/software-layer/pull/328, I also figured that it would be very useful to have something like this. The PRs had quite a few build issues, and multiple rebuilds were triggered in attempts to solve these issues, which makes it hard for a reviewer to determine the actual state of the builds/PR.

laraPPr:
I have created a `bot: status` command which collects all the comments in the pr related to the `bot: build` command which then print overveiw of all the build in one table. I'll open a pr for it once I have cleaned some things up.

laraPPr:
Just opened https://github.com/EESSI/eessi-bot-software-layer/pull/237 to create a `bot: status` command. 

also tested it and here is an example of what the command does:
- https://github.com/laraPPr/software-layer/pull/4#issuecomment-1910007732

trz42:
At least partially addressed with `bot: status` command.

PR #127, opened 2022-12-14
providing an overview of a PR's status in its description

trz42:
PR for coordinating work on #96 during Dec'22 hackathon

### DIFF
--- a/app.cfg.example
+++ b/app.cfg.example
@@ -120,3 +120,9 @@ poll_interval = 60
 
 # full path to the command for manipulating existing jobs
 scontrol_command = /usr/bin/scontrol
+
+[control_center]
+# designate this instance to manage the control center (collecting
+# information from other instances and updating the PR description;
+# with "PR description" we refer to the initial "comment" in a PR)
+manage_control_center = False
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -18,7 +18,7 @@ import waitress
 from connections import github
 from tools import config
 from tools.args import event_handler_parse
-from tools.controlcenter import control_center_process_comment
+from tools.controlcenter import control_center_process_comment, control_center_process_pr_opened
 from tasks.build import submit_build_jobs
 from tasks.deploy import deploy_built_artefacts
 
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -74,6 +74,9 @@ class EESSIBotSoftwareLayer(PyGHee):
         """
         Handle opening of a pull request.
         """
+        # if bot instance responsible for control center, process event
+        if False:
+            control_center_process_pr_opened(event_info, pr)
         log("PR opened: waiting for label bot:build")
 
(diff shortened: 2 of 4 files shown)


trz42:
Format for event log in PR comment. Using `<details>`
<details><summary>machine-readable event log</summary>
<p>
```python
[ {"timestamp": 1671051951, "action": submit, "result": 0, "details": {"jobid": 1234, "workdir": "/tmp"}}, {"timestamp": 1671052168, "action": release, "result": 0, "details": {}} ]
```
</p>
</details>

tools/controlcenter.py
```diff
@@ -0,0 +1,24 @@
...
+# EESSI software layer, see https://github.com/EESSI/software-layer
+#
+# author: Kenneth Hoste (@boegel)
+# author: Thomas Roeblitz (@trz42)
+#
+# license: GPLv2
+#
+
+
+def control_center_process_comment(request_body):
```
boegel, line 14: Maybe a shorter function name like `cc_process_comment` is better?

eessi_bot_event_handler.py
```diff
@@ -35,6 +36,9 @@ def handle_issue_comment_event(self, event_info, log_file=None):
         comment_author = request_body['comment']['user']['login']
         comment_txt = request_body['comment']['body']
         log("Comment posted in %s by @%s: %s" % (issue_url, comment_author, comment_txt))
+        # if bot instance responsible for control center, process event
+        if False:
```
boegel, line 40: I would do the `if` in the function itself, where we can also log a message that we got an event, but that we're not doing anything with it because this bot instance is not configured to do so?

tools/controlcenter.py
boegel: We'll need a separate function for new comments vs updated comments, so perhaps `cc_process_new_comment` and `cc_process_updated_comment`?

boegel (changes requested):

app.cfg.example
```diff
@@ -120,3 +120,9 @@ poll_interval = 60
 
 # full path to the command for manipulating existing jobs
 scontrol_command = /usr/bin/scontrol
+
+[control_center]
+# designate this instance to manage the control center (collecting
+# information from other instances and updating the PR description;
+# with "PR description" we refer to the initial "comment" in a PR)
+manage_control_center = False
```
boegel, line 128: Good enough for now, but I wonder if we should (later) just have a "`main_bot = True`", where being the main bot implies being responsible for managing the control center, but perhaps also other stuff. Just a thought.

trz42:
Can we distinguish new comments from updated comments?

boegel:
> Can we distinguish new comments from updated comments?

Based on https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=edited#issue_comment, yes: an event for a new comment will have `action` set to `created`, an event for an updated comment will have `edited` as `action` value.

trz42:
Very much stalled. Closing it for now.

PR #128, opened 2022-12-16
draft PR for generic read config

### LINKED ISSUE #112: Create a generic `read_config` function that reads the full `app.cfg`
From a comment to #62:

_"We should have a `read_config` function that reads the full deploy configuration, so here we can just use:

```python
cfg = read_config(DEPLOY_CFG)
deploy_permission = cfg[DEPLOY_PERMISSION]
if labeler not in deploy_permission:
```

`read_config` would then take care of the `.get` bit with a sensible default, the logging of the settings, and of the split for `deploy_permission`._

The same (taking care of split or similar) applies to other options with structured values, e.g., `cvmfs_customisations`.

PR #85 attempts to implement such a function and also defines a _required_ set of configuration options. Somehow the _constants_ (e.g., `DEPLOY_CFG` or `DEPLOY_PERMISSION`) had to be prefixed with the module name `config.` when they were used. Maybe there is a way to avoid that (just makes the code more lengthy).

Hafsa-Naeem:
Fixes #112

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -382,6 +382,29 @@ def determine_tarballs_to_deploy(successes, upload_policy):
     return to_be_deployed
 
 
+def read_config(cfg, option=None):
+    deploy_config = {}
+
+    funcname = sys._getframe().f_code.co_name
+
+    cfg = config.get_section(cfg)
+
+    log(f"{funcname}(): read configuration section '{cfg}' and option '{option}'")
+
+    if option is not None:
+        option_value = cfg.get(option)
+        log(f"{funcname}(): {option} '{option_value}'")
+
+        if " " in option_value:
+            deploy_config[option] = option_value.split()
+        else:
+            deploy_config[option] = option_value
+
+        return deploy_config[option]
+
+    return None
+
+
 def deploy_built_artefacts(pr, event_info):
     """Deploy built artefacts.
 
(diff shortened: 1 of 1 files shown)


tasks/deploy.py
```diff
@@ -382,6 +382,29 @@ def determine_tarballs_to_deploy(successes, upload_policy):
     return to_be_deployed
 
 
+def read_config(cfg, option=None):
```
trz42, line 385: Please add docstring for `read_config`

tasks/deploy.py
```diff
@@ -391,26 +414,20 @@ def deploy_built_artefacts(pr, event_info):
...
 
-    log(f"{funcname}(): deploy for PR {pr.number}")
-
-    deploycfg = config.get_section(DEPLOYCFG)
-
     # verify that the GH account that set label bot:deploy has the
     # permission to trigger the deployment
-    deploy_permission = deploycfg.get(DEPLOY_PERMISSION, '')
-    log(f"{funcname}(): deploy permission '{deploy_permission}'")
+    cfg = read_config(DEPLOYCFG, DEPLOY_PERMISSION)
```
boegel, line 419: `read_config` should be a function that is imported from `tools.config`, which reads a given configuration file at a specified path, and returns the configuration (as a dictionary). The idea is that this replaces the usage of the `read_file` (which is called in the `main` function in `eessi_bot_job_manager.py` and `eessi_bot_job_manager.py`) and `get_section` functions (both can be removed), as follows:

```python
cfg = read_config('app.cfg')
...
if labeler not in cfg[DEPLOYCFG][DEPLOY_PERMISSION]:
...
upload_policy = cfg[DEPLOYCFG][DEPLOY_PERMISSION]
```

That way, we can stop using the global variable `_config` in `tools/config.py`, which is initialized

boegel (changes requested):

tasks/deploy.py
boegel: or:

```python
cfg = read_config('app.cfg')
deploy_cfg = cfg[DEPLOYCFG]
...
if labeler not in deploy_cfg[DEPLOY_PERMISSION]:

trz42 (changes requested):

boegel:
replaced by #130

PR #130, opened 2022-12-21
generic read_config method in tools/config.py

Hafsa-Naeem:
see #112

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -117,7 +117,7 @@ class EESSIBotSoftwareLayer(PyGHee):
 def main():
     """Main function."""
     opts = event_handler_parse()
-    config.read_file("app.cfg")
+    config.read_config()
     github.connect()
 
     if opts.file:
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -117,6 +117,8 @@ class EESSIBotSoftwareLayer(PyGHee):
 def main():
     """Main function."""
     opts = event_handler_parse()
+
+    # config is read to raise an exception early when the event_handler starts.
     config.read_config()
     github.connect()
 
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -50,6 +50,9 @@ class EESSIBotSoftwareLayerJobManager:
     def __init__(self):
         self.logfile = os.path.join(os.getcwd(), "eessi_bot_job_manager.log")
 
+    def log(self, msg):
+        log(msg, log_file=self.logfile)
+
     def get_current_jobs(self):
         # who am i
         username = os.getenv('USER', None)
@@ -571,7 +574,7 @@ def main():
     """Main function."""
 
     opts = job_manager_parse()
-    config.read_file("app.cfg")
+    config.read_config()
     github.connect()
 
     job_manager = EESSIBotSoftwareLayerJobManager()
@@ -608,7 +611,8 @@ def main():
     poll_interval = 0
     job_manager.scontrol_command = ""
     if max_iter != 0:
-        job_mgr = config.get_section("job_manager")
+        cfg = config.read_config()
+        job_mgr = cfg["job_manager"]
         job_manager.job_ids_dir = job_mgr.get("job_ids_dir")
         job_manager.submitted_jobs_dir = os.path.join(
             job_manager.job_ids_dir, "submitted"
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -50,9 +50,6 @@ class EESSIBotSoftwareLayerJobManager:
     def __init__(self):
         self.logfile = os.path.join(os.getcwd(), "eessi_bot_job_manager.log")
 
-    def log(self, msg):
-        log(msg, log_file=self.logfile)
-
     def get_current_jobs(self):
         # who am i
         username = os.getenv('USER', None)
@@ -574,6 +571,8 @@ def main():
     """Main function."""
 
     opts = job_manager_parse()
+
+    # config is read to raise an exception early when the job_manager runs.
     config.read_config()
     github.connect()
 
--- a/tasks/build.py
+++ b/tasks/build.py
@@ -32,6 +32,7 @@ LOCAL_TMP = "local_tmp"
 SLURM_PARAMS = "slurm_params"
 SUBMIT_COMMAND = "submit_command"
 BUILD_PERMISSION = "build_permission"
+ARCHITECTURE_TARGETS = "architecturetargets"
 
 
 def get_build_env_cfg():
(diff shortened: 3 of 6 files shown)


tools/config.py
```diff
@@ -28,6 +29,24 @@ def read_file(path):
         error(f'Unable to read configuration file {path}!')
 
 
+def read_config(path):
```
boegel, line 32: @Hafsa-Naeem I would use `app.cfg` as default value for `path`, so we can just use `read_config()` instead of `read_config('app.cfg')`

```python
def read_config(path='app.cfg'):

File: tools/config.py [hunk 2]
```diff
@@ -28,6 +29,24 @@ def read_file(path):
...
+    """Read the config file
+    Args:
+        path (string): path to the configuration file
+    Returns:
+        dict (str, dict): dictionary containing configuration settings
+    """
+    fn = sys._getframe().f_code.co_name
+
+    config = configparser.ConfigParser()
```
boegel, line 41: please move this line into the `try` (as was done in `read_file`)

tasks/deploy.py
```diff
@@ -391,26 +391,21 @@ def deploy_built_artefacts(pr, event_info):
...
     funcname = sys._getframe().f_code.co_name
 
-    log(f"{funcname}(): deploy for PR {pr.number}")
-
-    deploycfg = config.get_section(DEPLOYCFG)
-
     # verify that the GH account that set label bot:deploy has the
     # permission to trigger the deployment
-    deploy_permission = deploycfg.get(DEPLOY_PERMISSION, '')
-    log(f"{funcname}(): deploy permission '{deploy_permission}'")
```
boegel, line 401: Let's keep this log message:

```python
cfg = config.read_config('app.cfg')
deploy_cfg = cfg[DEPLOYCFG]
deploy_permission = deploycfg.get(DEPLOY_PERMISSION, '')
log(f"{funcname}(): deploy permission '{deploy_permission}'")

File: tasks/deploy.py [hunk 4]
```diff
@@ -391,26 +391,21 @@ def deploy_built_artefacts(pr, event_info):
     """
     funcname = sys._getframe().f_code.co_name
 
-    log(f"{funcname}(): deploy for PR {pr.number}")
```
boegel, line 394: I don't see a reason to remove this log message?

tasks/deploy.py
```diff
@@ -391,26 +391,21 @@ def deploy_built_artefacts(pr, event_info):
...
-
-    deploycfg = config.get_section(DEPLOYCFG)
-
     # verify that the GH account that set label bot:deploy has the
     # permission to trigger the deployment
```
boegel, line 395: This comment should be moved down a bit, right above the `if labeler not in ...`

tasks/deploy.py
```diff
@@ -391,26 +391,21 @@ def deploy_built_artefacts(pr, event_info):
...
+    if labeler not in deploy_cfg[DEPLOY_PERMISSION]:
         log(f"{funcname}(): GH account '{labeler}' is not authorized to deploy")
         # TODO update PR comments for this bot instance?
         return
     else:
         log(f"{funcname}(): GH account '{labeler}' is authorized to deploy")
 
     # get upload policy from config
-    upload_policy = deploycfg.get(UPLOAD_POLICY)
-    log(f"{funcname}(): upload policy '{upload_policy}'")
```
boegel, line 413: Also keep this log message

tasks/build.py
```diff
@@ -21,6 +21,7 @@
 from pyghee.utils import log, error
 from tools import config, run_cmd
 
+BUILDENV = "buildenv"
```
boegel, line 24: Also actually uses this constant below:

```python
build_cfg = cfg[BUILDENV]

File: tasks/build.py [hunk 8]
```diff
@@ -99,8 +101,10 @@ def get_architecturetargets():
         dict(str, dict): dictionary of arch_target_map which contains entries of the format
                          OS/SUBDIR : ADDITIONAL_SBATCH_PARAMETERS
     """
-    architecturetargets = config.get_section('architecturetargets')
-    arch_target_map = json.loads(architecturetargets.get('arch_target_map'))
+    cfg = config.read_config('app.cfg')
+    architecturetargets_cfg = cfg['buildenv']
+    architecturetargets = architecturetargets_cfg['architecturetargets']
```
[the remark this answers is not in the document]
boegel: This line can be removed, since it's not used at all...

There's actually no such setting at all, so without using `.get` this will trigger a hard crash!

tasks/build.py
```diff
@@ -395,16 +401,17 @@ def check_build_permission(pr, event_info):
...
     # verify that the GH account that set label bot:build has the
     # permission to trigger the build
-    build_permission = buildenv.get(BUILD_PERMISSION, '')
+    build_permission = build_cfg[BUILD_PERMISSION]
 
     log(f"{funcname}(): build permission '{build_permission}'")
 
     build_labeler = event_info['raw_request_body']['sender']['login']
-    if build_labeler not in build_permission.split():
+    if build_labeler not in build_cfg[BUILD_PERMISSION]:
```
boegel, line 414: Use the `build_permission` defined above, and keep the `.split()`

tasks/deploy.py
```diff
@@ -391,26 +391,21 @@ def deploy_built_artefacts(pr, event_info):
...
     # verify that the GH account that set label bot:deploy has the
     # permission to trigger the deployment
-    deploy_permission = deploycfg.get(DEPLOY_PERMISSION, '')
-    log(f"{funcname}(): deploy permission '{deploy_permission}'")
+    cfg = config.read_config('app.cfg')
+    deploy_cfg = cfg[DEPLOYCFG]
 
     labeler = event_info['raw_request_body']['sender']['login']
-    if labeler not in deploy_permission.split():
+    if labeler not in deploy_cfg[DEPLOY_PERMISSION]:
```
boegel, line 400: keep `deploy_permission.split()` (especially the `.split()` part is important)

boegel (changes requested):
@Hafsa-Naeem You should also *remove* the `read_file` and `get_section` functions, after you make sure they're no longer used.

In general, it's important to keep using `.get` when grabbing a particular configuration setting, otherwise the bot will crash if a particular setting is not provided in the configuration file:
```python
>>> example_cfg = {}
>>> print(example_cfg.get('example'))
None
>>> print(example_cfg['example'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'example'
```

tasks/build.py
```diff
@@ -99,8 +101,10 @@ def get_architecturetargets():
         dict(str, dict): dictionary of arch_target_map which contains entries of the format
                          OS/SUBDIR : ADDITIONAL_SBATCH_PARAMETERS
     """
-    architecturetargets = config.get_section('architecturetargets')
-    arch_target_map = json.loads(architecturetargets.get('arch_target_map'))
+    cfg = config.read_config('app.cfg')
+    architecturetargets_cfg = cfg['buildenv']
```
boegel, line 105:
```python
architecturetargets_cfg = cfg['architecturetargets']

**Review by boegel — CHANGES_REQUESTED** [review]

File: tasks/build.py [hunk 12]
```diff
@@ -355,7 +359,9 @@ def submit_build_jobs(pr, event_info):
     """
     # retrieving some settings from 'app.cfg' in bot directory
     # [github]
-    app_name = config.get_section('github').get('app_name')
+    cfg = config.read_config('app.cfg')
```
boegel, line 362: This can be changed to
```python
cfg = config.read_config()
```

if `read_config` is reading `app.cfg` by default

boegel (dismissed):

eessi_bot_job_manager.py
```diff
@@ -50,6 +50,9 @@ class EESSIBotSoftwareLayerJobManager:
     def __init__(self):
         self.logfile = os.path.join(os.getcwd(), "eessi_bot_job_manager.log")
 
+    def log(self, msg):
+        log(msg, log_file=self.logfile)
+
```
trz42, line 55: Maybe good to check if all uses of

```python
log(..., log_file=self.logfile)
```

are replaced by the simpler call

```python
log(...)
```

trz42 (commented):
Overall looks good to me. A few possible improvements, but those are not blocking the merge.

However, the configuration file is now read rather often compared to only once before. Besides the minor performance hit, this may cause unwanted / difficult to debug side effects when the configuration file is changed while the bot is running. That might be fixed by re-introducing a global variable storing the configuration and only reading it once at the beginning.

eessi_bot_event_handler.py
```diff
@@ -117,7 +117,7 @@ def start(self, app, port=3000):
 def main():
     """Main function."""
     opts = event_handler_parse()
-    config.read_file("app.cfg")
+    config.read_config()
```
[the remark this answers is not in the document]
trz42: Suggest to add a comment that config is read to raise an exception early when the event handler starts.

eessi_bot_job_manager.py
trz42: Maybe it leaked into this PR from PR #131 ?

trz42 (approved):
Looks good. Thanks @Hafsa-Naeem!

PR #131, opened 2022-12-27
dedicated log

Hafsa-Naeem:


tools/config.py
```diff
@@ -28,6 +29,24 @@ def read_file(path):
...
+
+    try:
+        config = configparser.ConfigParser()
+        config.read(path)
+    except Exception:
+        error(f"{fn}(): Unable to read configuration file {path}!")
+
+    return config
+
+
```
trz42, line 49: Has this leaked into the PR from #130?

app.cfg.example
```diff
@@ -111,10 +111,16 @@ deploy_permission = trz42
 #   and what submission parameters shall be used
 arch_target_map = { "linux/x86_64/generic" : "--constraint shape=c4.2xlarge", "linux/x86_64/amd/zen2": "--constraint shape=c5a.2xlarge" }
 
+[event_handler]
+# configuration for event handler which receives events from a GitHub repository.
+log_path = /path/to/eessi_bot_event_handler.log
```
trz42, line 116: The other sections don't yet have a comment describing their purpose. So adding one here is good. However, would be better to move the comment `# configuration for ...` before `[event_handler]`.

For `log_path` there should be a short comment describing what this option is used for.

app.cfg.example
```diff
@@ -111,10 +111,16 @@ deploy_permission = trz42
...
+
 
 [job_manager]
 # directory where job manager stores information about jobs to be tracked
 #   e.g. as symbolic link JOBID -> directory to job
+log_path = /path/to/eessi_bot_job_manager.log
+
 job_ids_dir = $HOME/jobs/ids
```
trz42, line 124: The new option `log_path` and a short comment describing its purpose should be placed before or after the option `job_ids_dir` and its describing comment. Putting it in between makes it difficult to read if not wrong.

trz42 (dismissed):
A few comments/questions should be addressed.

A short section in `README.md` describing the new configuration options should be added.

tools/config.py
trz42: After #130 is merged pull in changes into development branch (dedicated_log). Maybe conflicts need to be resolved.

eessi_bot_event_handler.py
```diff
@@ -23,10 +24,25 @@
 
 from pyghee.lib import PyGHee, create_app, get_event_info, read_event_from_json
 from pyghee.utils import log
-import os
 
 
 class EESSIBotSoftwareLayer(PyGHee):
+    def __init__(self):
+        super().__init__()
```
boegel, line 31: Better/safer (maybe needs a better docstring):

```python
    def __init__(self, *args, **kwargs):
        """Constructor."""
        super(EESSIBotSoftwareLayer, self).__init__(*args, **kwargs)
```

eessi_bot_event_handler.py
```diff
@@ -23,10 +24,25 @@
...
+        cfg = config.read_config()
```
boegel, line 32: Maybe we can use `self.cfg` here, and avoid having to read the configuration again later, just use `self.cfg` instead?

eessi_bot_event_handler.py
```diff
@@ -23,10 +24,25 @@
...
+        self.logfile = event_handler_cfg.get('log_path')
+
+    def log(self, msg):
+        """this log method will pass values (message and log file location) to PYGHEE log method
+
+        Args:
+            msg (string): message to log in the eessi_bot_event_handler.log
+        """
+        funcname = sys._getframe().f_back.f_code.co_name
+        msg = "the method: %s logged: %s" % (funcname, msg)
```
boegel, line 43: I would reformat this to:

```python
msg = "[%s] %s" % (funcname, msg)

File: eessi_bot_event_handler.py [hunk 7]
```diff
@@ -23,10 +24,25 @@
...
 
 
 class EESSIBotSoftwareLayer(PyGHee):
+    def __init__(self):
+        super().__init__()
+        cfg = config.read_config()
+        event_handler_cfg = cfg['event_handler']
+        self.logfile = event_handler_cfg.get('log_path')
+
+    def log(self, msg):
```
boegel, line 36: Would be nicer to support "lazy logging", where formatting is only done when logging is done:

```python
def log(self, msg, *args):
    """..."""
    if args:
        msg = msg % args
```

then you can use:

```python
self.log("example: %s (%s)", foo, bar)
```

instead of having to pre-format the message:

```python
self.log("example: %s (%s)" % (foo, bar))

**Review by boegel — DISMISSED** [review]

File: eessi_bot_event_handler.py (hunk 5 above)
Comment (boegel) [reply eessi_bot_event_handler.py:32]: This would actually fix #134

**Comment by trz42** [comment]
Tested PR on AWS cluster and eX3 cluster running the bot:build step. See https://github.com/trz42/software-layer/pull/50#issuecomment-1375641543

With that the PR is ready to get merged.

**Review by trz42 — APPROVED** [review]
All requested changes have been implemented. The updated PR has been test with a real pull request.

File: eessi_bot_job_manager.py [hunk 8]
```diff
@@ -48,7 +48,9 @@ class EESSIBotSoftwareLayerJobManager:
     "main class for (Slurm) job manager of EESSI bot (separate process)"
 
     def __init__(self):
-        self.logfile = os.path.join(os.getcwd(), "eessi_bot_job_manager.log")
+        cfg = config.read_config()
+        job_manager_cfg = cfg['job_manager']
+        self.logfile = job_manager_cfg.get('log_path')
```
boegel, line 53: @trz42 Would be nice to also add a `log` method that uses `self.logfile` in `eessi_bot_job_manager.py`, no?

PR #132, opened 2022-12-27
retry after exception

Hafsa-Naeem:
addresses one case of #20

### DIFF
--- a/connections/github.py
+++ b/connections/github.py
@@ -10,6 +10,8 @@
 # license: GPLv2
 #
 import datetime
+import time
+
 
 from tools import config, logging
 from github import Github, GithubIntegration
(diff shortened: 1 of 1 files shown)


connections/github.py
```diff
@@ -28,15 +30,23 @@ def get_token():
...
+        # If the config keys are not set, get_access_token will raise a NotImplementedError
+        # Returning NoneType token will stop the connection in get_instance
+        try:
+            github_integration = GithubIntegration(app_id, private_key)
+            # Note that installation access tokens last only for 1 hour,
+            # you will need to regenerate them after they expire.
+            _token = github_integration.get_access_token(installation_id)
+        except NotImplementedError as err:
+            if i < tries - 1:
+                time.sleep(0.8)
```
trz42, line 44: This always waits 0.8 seconds. Maybe the waiting time could be $n * (i+1)$ seconds? $n$ being the default/minimum waiting time (thus waiting time would increase linearly: $n, 2n, 3n, \dots$). Alternatively, the waiting time could be $n^{(i+1)}$ seconds (thus waiting time would increase exponentially: $n, n^2, n^3, \dots$).

connections/github.py
```diff
@@ -28,15 +30,23 @@ def get_token():
...
+                continue
+            else:
+                logging.error(err)
+                _token = None
+        break
```
trz42, line 49: Is the `break` needed?

trz42 (changes requested):
Looks good. Some minor modifications suggested.

connections/github.py
trz42: To be on the safe side, maybe wait for $1 + i * n$ seconds.

connections/github.py
trz42: Yes it's needed. Probably easier to read if the `break` is moved to after line 41.

trz42 (approved):

PR #141, opened 2023-01-15
let bot pass CPU target to build for to build job script, and use it to set $EESSI_SOFTWARE_SUBDIR_OVERRIDE

### LINKED ISSUE #140: bot build script should make sure correct CPU target is used
The install script for the EESSI software layer currently relies on `archspec` for detecting for which part of the software layer the software installations should be created.

For the most recent CPUs (`aarch64/graviton3` and `x86_64/amd/zen3`, currently), that doesn't work as intended though, because the `archspec` included with the EESSI 2021.12 pilot (v0.1.3) doesn't correctly recognize these yet (it reports `graviton2` and `zen2`, respectively, cfr. https://github.com/archspec/archspec-json/issues/38 + https://github.com/archspec/archspec-json/issues/39).

During the build, the bot should ensure that the correct subdirectory is used for the builds.

That can be done by defining `$EESSI_SOFTWARE_SUBDIR_OVERRIDE`, but the [`EESSI-pilot-install-software.sh`](https://github.com/EESSI/software-layer/blob/main/EESSI-pilot-install-software.sh) script should also be updated to not blindly overwrite that environment variable, like it does now...

boegel:
fixes #140

### DIFF
--- a/scripts/eessi-bot-build.slurm
+++ b/scripts/eessi-bot-build.slurm
@@ -131,6 +131,11 @@ fi
 
 source init/minimal_eessi_env  # for $EESSI_PILOT_VERSION and $EESSI_OS_TYPE
 
+# $CPU_TARGET is set by event handler when job is submitted;
+# $EESSI_SOFTWARE_SUBDIR_OVERRIDE is picked up by eessi_software_subdir.py,
+# and is also passed to the install script by build_container.sh & co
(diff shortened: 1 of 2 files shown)


trz42 (approved):
Looks good. Tested with [EESSI/software-layer PR#221](https://github.com/EESSI/software-layer/pull/221) on two systems. Results were a bit mixed, but it should be ok (cfr. https://github.com/trz42/software-layer/pull/53).

PR #144, opened 2023-01-19
run bot/build.sh script in build job script, if it exists

boegel:
No changes to the code in the `else` part, keeping that as fallback now so we can keep building PRs even if no `bot/build.sh` script exists yet...


trz42 (approved):
Looks good. Tested on one system.

Issue #150, opened 2023-02-08
Add CI test before fixing bug in determining running jobs


trz42:
Issue #133 reports a bug in the code that determines running jobs.

Before a PR for #133 is submitted a test should be written. The goal with that test is to ensure to capture changes in the code.

trz42:
Fixed by #153

Issue #133, opened 2022-12-28
job manager reports non-running jobs as running


trz42:
It seems that the job manager reports running jobs too early. See for example https://github.com/trz42/software-layer/pull/50#issuecomment-1366882961

The relevant code in `eessi_bot_job_manager.py` is

```python
    def determine_running_jobs(self, known_jobs, current_jobs):
        """ determine which jobs are in running state
        Args:
            known_jobs (dict): dictionary containing data of known jobs
            current_jobs (dict): dictionary containing data of current jobs
        Returns:
            running_jobs (dict): dictionary containing data of ruuning jobs
        """
        running_jobs = []
        for rkey in current_jobs:
            if rkey in known_jobs:
                running_jobs.append(rkey)

        return running_jobs
```

This returns all jobs as _running_ that are known (from one iteration ago) and current (from the current iteration). Jobs could be still pending or, as in the above example PR, waiting until the compute node gets launched.

The above method would also need to check if the status of job `rkey` is actually `RUNNING`.

trz42:
Fixed by #153

PR #153, opened 2023-02-22
Fix for bug in determine_running_jobs in job_manager

jonas-lq:
Fix for issue #133 and related tests as described in #150


trz42 (commented):
Looks good to me. Would like to run a test before approving.

Good work @jonas-lq !

trz42 (approved):
Changes look good. Tests pass. Also tested with a PR on AWS (see https://github.com/trz42/software-layer/pull/58). Works as intended.

Issue #107, opened 2022-12-06
Rename config upload_to_s3_script


trz42:
From #62:

_"Perhaps rename to `tarball_upload_script`? The S3 aspect may change at some point, or perhaps the script starts accepting options for where to upload to, etc."_

trz42:
Closed by #154

PR #154, opened 2023-02-24
Renamed upload_to_s3_script to tarball_upload_script

jonas-lq:
Fix for issue #107


trz42 (commented):
Looks good to me. Will make test run, likely not before early March. Then we can merge it. Should be no problem as the changes are small and isolated.

trz42 (approved):
Tested on AWS. See https://github.com/trz42/software-layer/pull/58#issuecomment-1455824298

PR #155, opened 2023-02-25
support for `bot/build.sh` in software layer

trz42:
Draft version of bot-side code to support `bot/build.sh` in software layer (see https://github.com/EESSI/software-layer/pull/233).

- stores job config data in `WorkDir/cfg/job.cfg`
- makes use of `--repository` capability of `eessi_container.sh`
- reports configured targets and repositories when PR is opened
- adds information about target repository to PR comment when a job is submitted
- adds support for `container_cachedir` setting in `app.cfg` to reuse cache across jobs (speeds up job start)

~Still draft because~ Ready to get reviewed after tasks below are done:
- [x] more testing should be done (while a similar version is in use for NESSI for ~2 weeks + this exact version has been tested with a few jobs ... more testing is beneficial)
  - NESSI bot instances have been updated and bots work as before (full cycle, building + ingesting, done once + more build jobs ran)
- [x] documentation for defining repositories needs to be added
  - maybe good enough if specifying `EESSI-pilot` works (make a test) ... works after fixing some issues
  - if it works, create an issue to add more documentation ... created issue https://github.com/EESSI/eessi-bot-software-layer/issues/166
- [x] CREATE an issue (see #162) for this and do the change in a follow-up PR
  - currently jobs are generated while iterating over `arch_target_map` entries, then determining the repository via `repo_target_map` ... it should be the other way around ... also if there is no matching entry in `repo_target_map` or `arch_target_map` for a given architecture an error should be logged and the bot shall continue (currently results in a crash)


boegel (changes requested):
partial review (didn't check changes to `tasks/build.py` yet)

tasks/build.py
```diff
@@ -19,11 +19,14 @@
 from collections import namedtuple
 from connections import github
 from datetime import datetime, timezone
+from distutils.dir_util import copy_tree
```
boegel, line 22: OK to use for now, but `distutils` is deprecated since Python 3.10, and will be removed in Python 3.12, so we should open an issue to follow up and make sure we don't rely on `distutils` at all going forward.

tasks/build.py
```diff
@@ -206,48 +318,145 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
+    log(f"{fn}(): {len(jobs)} jobs to proceed after applying white list")
     if jobs:
         log(json.dumps(jobs, indent=4))
 
-    return repo_name, jobs
+    return jobs
+
+
+def prepare_job_cfg(job_dir, build_env_cfg, repos_cfg, repo_id, software_subdir, os_type):
+    """setup cfg/job.cfg
```
boegel, line 381:
```suggestion
    """
    Set up job configuration file 'cfg/job.cfg'
```

tasks/build.py
```diff
@@ -206,48 +318,145 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
+        build_env_cfg (dictionary): build environment configuration
+        repos_cfg (dictionary):  configuration settings for all repositories
+        repo_id (string):  identifier of the repository to build for
+        software_subdir (string): software subdirectory to build for (CPU arch)
+        os_type (string): type of the os (e.g., linux)
+    """
+    fn = sys._getframe().f_code.co_name
+
+    jobcfg_dir = os.path.join(job_dir, 'cfg')
+    os.makedirs(jobcfg_dir, exist_ok=True)
```
boegel, line 394: another place where we could use `try`/`except` around `makedirs`, which tells me we should have our own `mkdir` function (let's open an issue on that, it's just minor code cleanup)

tasks/build.py
```diff
@@ -206,48 +318,145 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
+    job_cfg[JOB_ARCHITECTURE] = {}
+    job_cfg[JOB_ARCHITECTURE][JOB_SOFTWARE_SUBDIR] = software_subdir
+    job_cfg[JOB_ARCHITECTURE][JOB_OS_TYPE] = os_type
+
+    config_data = StringIO()
+    job_cfg.write(config_data)
+
+    jobcfg_file = os.path.join(jobcfg_dir, 'job.cfg')
+    with open(jobcfg_file, "w") as jcf:
+        job_cfg.write(jcf)
```
boegel, line 450: This looks wrong to me, should be `jfc.write(...)`?

boegel (changes requested):

tasks/build.py
```diff
@@ -206,48 +318,145 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
-        job = Job(arch_job_dir, arch_target, slurm_opt)
-        jobs.append(job)
-    log("  %d jobs to proceed after applying white list" % len(jobs))
+    jobs = []
+    for arch, slurm_opt in arch_map.items():
+        # TODO check if arch in repos_map
+        arch_dir = arch.replace('/', '_')
+        # check if repo_target_map contains an entry for {arch}
+        if arch not in repocfg[REPO_TARGET_MAP]:
+            log(f"{fn}(): skipping arch {arch} because repo target map does not define repositories to build for")
```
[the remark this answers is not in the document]
boegel: Just logging is OK, we can make sure configuration is sane on startup in another PR

tasks/build.py
```diff
@@ -206,48 +318,145 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
+            continue
+        for repo_id in repocfg[REPO_TARGET_MAP][arch]:
+            # ensure repocfg contains information about the repository repo_id if repo_id != EESSI-pilot
+            # Note, EESSI-pilot is a bad/misleading name, it should be more like AS_IN_CONTAINER
+            if repo_id != "EESSI-pilot" and repo_id not in repocfg:
+                log(f"{fn}(): skipping repo {repo_id}, it is not defined in repo config {repocfg[REPOS_CFG_DIR]}")
```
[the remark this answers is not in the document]
boegel: Just logging is OK, we can make sure configuration is sane on startup in another PR

boegel (approved):

Issue #146, opened 2023-01-25
submit build jobs with default time limit, add configuration option to use different time limit


boegel:
By default, the bot now submits build jobs without any time limit, which is a problem in case the build job "hangs" for some reason.

We should use a default time limit (24h?), and add a configuration option to use a different time limit.

trz42:
Closed by #156

PR #156, opened 2023-02-28
Added 24h default time limit

jonas-lq:
Fix for issue #146

### DIFF
--- a/tasks/build.py
+++ b/tasks/build.py
@@ -267,7 +267,7 @@ def submit_job(job, submitted_jobs, build_env_cfg, ym, pr_id):
 
     # Add a default time limit of 24h to the command if nothing else is specified by the user
     if "--time=" not in build_env_cfg[SLURM_PARAMS]:
-        time_limit= "--time=24:00:00"
+        time_limit = "--time=24:00:00"
     else:
         time_limit = ""
 
--- a/tasks/build.py
+++ b/tasks/build.py
@@ -24,7 +24,7 @@ from tools import config, run_cmd
 
 BUILDENV = "buildenv"
 BUILD_JOB_SCRIPT = "build_job_script"
-DEFAULT_JOB_TIME_LIMIT="24:00:00"
+DEFAULT_JOB_TIME_LIMIT = "24:00:00"
 CVMFS_CUSTOMIZATIONS = "cvmfs_customizations"
 HTTP_PROXY = "http_proxy"
 HTTPS_PROXY = "https_proxy"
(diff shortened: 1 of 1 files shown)


tasks/build.py
```diff
@@ -264,13 +265,20 @@ def submit_job(job, submitted_jobs, build_env_cfg, ym, pr_id):
             - job_id(string):  job_id of submitted job
             - symlink(string): symlink from main pr_<ID> dir to job dir (job[0])
     """
-
     # determine CPU target to use: job.arch_target, but without the linux/ part
     cpu_target = '/'.join(job.arch_target.split('/')[1:])
 
+    # Add a default time limit of 24h to the command if nothing else is specified by the user
+    all_opts = " ".join([build_env_cfg[SLURM_PARAMS], job.slurm_opts, '--export=ALL,CPU_TARGET=%s' % cpu_target])
+    if ("--time" in all_opts) or ("-t" in all_opts):
```
trz42, line 273: This is a big improvement. However, there are a few cases this way of checking is not catching. See test cases 8 & 9 in https://github.com/trz42/software-layer/pull/58#issuecomment-1458652800

So, we may need to iterate over all options and check if they begin with `--time` or `-t`, e.g.,
```python
opt.startswith("--time")
```

trz42 (changes requested):
Much better. Some not so unusual cases we should catch. See comment.

tasks/build.py
```diff
@@ -269,8 +269,9 @@ def submit_job(job, submitted_jobs, build_env_cfg, ym, pr_id):
     cpu_target = '/'.join(job.arch_target.split('/')[1:])
 
     # Add a default time limit of 24h to the command if nothing else is specified by the user
-    all_opts = " ".join([build_env_cfg[SLURM_PARAMS], job.slurm_opts, '--export=ALL,CPU_TARGET=%s' % cpu_target])
-    if ("--time" in all_opts) or ("-t" in all_opts):
+    all_opts_str = " ".join([build_env_cfg[SLURM_PARAMS], job.slurm_opts, '--export=ALL,CPU_TARGET=%s' % cpu_target])
```
trz42, line 272: I would remove the third argument `, '--export=ALL,CPU_TARGET=%' % cpu_target]`. This is not a configuration parameter that we expect to include a time limit. In a future version of the bot this code is probably going away anyway.

trz42 (changes requested):
Looks good. All test cases that failed before succeed now. Repeated some tests for the other cases as well, see https://github.com/trz42/software-layer/pull/58#issuecomment-1463402641

One final suggestion to polish a bit the code. Then it can be merged.

Issue #121, opened 2022-12-09
Add more information to log when labeler has no permission


trz42:
See https://github.com/EESSI/eessi-bot-software-layer/pull/62#discussion_r1023929322

PR #171, opened 2023-03-28
Bot comments when labels are set without permission

jonas-lq:
Fix for issue #121 

A comment gets added to the PR when the user sets one of the following labels without permission:
-bot:build
-bot:deploy


trz42 (dismissed):
This looks good. Tested it (see https://github.com/trz42/software-layer/pull/62). It works.

It only remains to sync the PR with the main branch in EESSI/eessi-bot-software-layer. Then it can be merged.

Nice work!

app.cfg.example
```diff
@@ -72,6 +72,7 @@ submit_command = /usr/bin/sbatch
 # if value is left/empty everyone can trigger the build
 # value can be a space delimited list of GH accounts
 build_permission = Hafsa-Naeem
+no_build_permission_comment = Label `bot:build` has been set by user `{build_labeler}`, but only users `{build_permission_users}` have permission to trigger the action
```
boegel, line 75: Hmm, not sure if we should let the bot report who they should social engineer into adding labels...

The message should just say

```
Label `bot:build` has been set by user `{build_labeler}`, but this person does not have permission to trigger builds
```

(and similar for deploy, of course)

boegel (changes requested):

tasks/build.py
```diff
@@ -440,7 +441,15 @@ def check_build_permission(pr, event_info):
...
     if build_labeler not in build_permission.split():
         log(f"{funcname}(): GH account '{build_labeler}' is not authorized to build")
-        # TODO update PR comments for this bot instance?
+        no_build_permission_comment = buildenv.get(NO_BUILD_PERMISSION_COMMENT)
+        gh = github.get_instance()
+        repo_name = event_info["raw_request_body"]["repository"]["full_name"]
+        repo = gh.get_repo(repo_name)
+        pull_request = repo.get_pull(pr.number)
+        pull_request.create_issue_comment(
+            no_build_permission_comment.format(build_labeler=build_labeler,
```
boegel, line 450: This code is almost identical to what we have for `deploy`, should this be factored out to a function used in both?

boegel (changes requested):

app.cfg.example
trz42: Well, particularly this was suggested/requested by you, no? See https://github.com/EESSI/eessi-bot-software-layer/pull/62#discussion_r1023929322

tasks/build.py
trz42: Yeah, can be done in some code polishing/refactoring.

bedroge: Created issue #180 for this.

bedroge (approved):
Lgtm

Issue #110, opened 2022-12-06
Automatically print function name when logging information


trz42:
PR #62 introduced the adding of the function name to logging information with the following code (pattern)

```python
import sys
...
funcname = sys._getframe().f_code.co_name
...
log(f"{funcname}(): deploy for PR {pr.number}")
```

It should be possible to move this into the `log` function, so the extra code is not needed and the information is always added. See, for example, https://stackoverflow.com/questions/5067604/determine-function-name-from-within-that-function-without-using-traceback for a good discussion of options (and their performance). The best is probably to test a view options. Also, having defined tests might be a good thing to start with.

edit: this depends on #125, so we have a central place to do this

Issue #91, opened 2022-12-03
Improve logging for bot components


trz42:
Logging could implement different levels, channels, include the current function name and be duplicated to stdout/stderr on request (e.g., in debugging sessions).

- For levels [filesystem-layer/PR#90](https://github.com/EESSI/filesystem-layer/pull/90) may be worth to look into.
- For channels, the idea would be to have a different means to switch on/off topical debugging (e.g., only for the build step, only for the deploy step, etc.)
- For always adding the function name, the `tasks/deploy.py` may be a good starting point. Also, https://stackoverflow.com/questions/5067604/determine-function-name-from-within-that-function-without-using-traceback may be a good resource to look into.

Also, it could be made more transparent to what file information is logged and to set this just once (at the start up) without the need to repeatedly adding the name of a `log_file`.

boegel:
One aspect of this is to avoid that we have two separate logs for event handler (`event_handler.log` and `pyghee.log`).

We should also have a single `log` function that makes sure that all log messages have the same format, are timestamped, mention where the log messages was coming from, etc.

We could also try and make both components log to a single place, which would require locking the log file while it's being written to (again a reason to have a single log file that we use everywhere).

PR #172, opened 2023-04-06
support for sending commands to bot instances via PR comments

trz42:
This PR probably needs some updates/polishing but it has been tested already with a few PRs to the software-layer repository. It includes the following additions:

- class to model action filters where an action is something like submitting a build job
  - compared to setting a label `bot:build` an action can be filtered wrt a repository, an architecture and a bot instance name
  - idea is that an action is only performed if its context (repository, architecture and bot instance) matches the filter
- class to model bot commands which are composed of a command such as `help` and `build` and an action filter
- adds code to parse *NEW* comments for bot commands (beginning at the start of a line with `bot: `), displays if a command has been received, processes the command(s) and adds some result/output in the same comment
- handlers for commands: `help`, `build` and `showconfig`
- adds an `app.cfg` setting for limiting which GH accounts can send commands to the bot
- refactors existing code to work well with the addition of action filters: one major change is how configuration settings are passed to where they are accessed
- includes some polishing of docstrings and other things
- last but not least, adds some unit tests for the action filters

~TODOs~ done
- addressed the below by only running the critical code if the event is an issue_comment ... then the request_body has the key 'issue'
  - clarify that `update_pr_comment` is working only with issue comment events (not with pull request events) because it accesses the `issue` section of an event, see line (in `tools/pr_comments.py`)
    ```python
    pr_number = int(request_body['issue']['number'])
    ```
  - or change the code such that it works with both pull request events and issue comment events, e.g., by determining the type of the event

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -125,7 +125,7 @@ class EESSIBotSoftwareLayer(PyGHee):
                     continue
                 commands.append(ebc)
                 self.log(f"found bot command: '{bot_command}'")
-                comment_response += "\n- received bot command `{bot_command}`"
+                comment_response += f"\n- received bot command `{bot_command}`"
                 comment_response += f" from `{sender}`"
                 comment_response += f"\n  - expanded format: `{ebc.to_string()}`"
             # FIXME add an else branch that logs information for comments not
@@ -170,13 +170,14 @@ class EESSIBotSoftwareLayer(PyGHee):
         for cmd in commands:
             try:
                 update = self.handle_bot_command(event_info, cmd)
-                comment_result += f"\n- handling `{cmd.command}` resulted in: "
-                comment_result += update
-                self.log(f"handling '{cmd.command}' resulted in '{update}'")
+                comment_result += f"\n- handling command `{cmd.to_string()}` resulted in: "
+                comment_result += f"\n  - {update}"
+                self.log(f"handling command '{cmd.to_string()}' resulted in '{update}'")
 
-            except EESSIBotCommandError as bce:
-                self.log(f"ERROR: handling {cmd.command} failed with {bce.args}")
-                comment_result += f"\n- handling `{cmd.command}` failed with {bce.args}"
+            except EESSIBotCommandError as err:
+                self.log(f"ERROR: handling command {cmd.command} failed with {err.args[0]}")
+                comment_result += f"\n- handling command `{cmd.command}` failed with message"
+                comment_result += f"\n  _{err.args[0]}_"
                 continue
             except Exception as err:
                 log(f"Unexpected err={err}, type(err)={type(err)}")
@@ -296,7 +297,7 @@ class EESSIBotSoftwareLayer(PyGHee):
             return handler(event_info, bot_command)
         else:
             self.log(f"No handler for command '{cmd}'")
-            raise EESSIBotCommandError(f"unknown command '{cmd}'; use `bot: help` for usage information")
+            raise EESSIBotCommandError(f"unknown command `{cmd}`; use `bot: help` for usage information")
 
     def handle_bot_command_help(self, event_info, bot_command):
         """handles command 'bot: help' with a simple usage info"""
@@ -332,7 +333,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         pr_number = event_info['raw_request_body']['issue']['number']
         pr = gh.get_repo(repo_name).get_pull(pr_number)
         issue_comment = self.handle_pull_request_opened_event(event_info, pr)
-        return f"added comment {issue_comment.issue_url} to show configuration"
+        return f"added comment {issue_comment.html_url} to show configuration"
 
     def start(self, app, port=3000):
         """starts the app and log information in the log file
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -176,15 +176,17 @@ class EESSIBotSoftwareLayer(PyGHee):
         for cmd in commands:
             try:
                 update = self.handle_bot_command(event_info, cmd)
-                # next two lines commented out to make the bot less noisy,
-                # could be enabled again if bot commands accept arg --verbose
-                # or bot instance has a similar setting; those were also a bit
-                # confusing when different instances overwrote comments of the
-                # other instances
-                # comment_update += f"\n- handling `{cmd.command}` resulted in: "
-                # comment_update += update
-                # update_pr_comment(event_info, comment_update)
+                # the update to the PR comment is only done if the command is
+                # 'help'; otherwise there was too much noise and also updates
+                # were a bit confusing when different instances overwrote
+                # comments of the other instances
+                if cmd.command == 'help':
+                    comment_update += f"\n- handling `{cmd.command}` resulted in: "
+                    comment_update += update
+                    update_pr_comment(event_info, comment_update)
+
                 self.log(f"handling '{cmd.command}' resulted in '{update}'")
+
             except EESSIBotCommandError as bce:
                 self.log(f"ERROR: handling {cmd.command} failed with {bce.args}")
                 # next two lines commented out to make the bot less noisy,
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -74,9 +74,17 @@ class EESSIBotSoftwareLayer(PyGHee):
         action = request_body['action']
         sender = request_body['sender']['login']
         owner = request_body['comment']['user']['login']
+        repo_name = request_body['repository']['full_name']
+        pr_number = request_body['issue']['number']
+
         # FIXME add request body text (['comment']['body']) to log message when
         # log level is set to debug
         self.log(f"Comment in {issue_url} (owned by @{owner}) {action} by @{sender}")
+
+        # obtain app name and format for reporting about received & processed
+        # commands
+        app_name = self.cfg[GITHUB][APP_NAME]
+        command_response_fmt = self.cfg[BOT_CONTROL][COMMAND_RESPONSE_FMT]
 
         # currently, only commands in new comments are supported
         #  - commands have the syntax 'bot: COMMAND [ARGS*]'
@@ -95,6 +103,13 @@ class EESSIBotSoftwareLayer(PyGHee):
         #      run the update through the function checking for a bot command.
         if check_command_permission(sender) is False:
             self.log(f"account `{sender}` has NO permission to send commands to bot")
+            comment_response = f"\n- account `{sender}` has NO permission to send commands to the bot"
+            comment_body = command_response_fmt.format(
+                app_name=app_name,
+                comment_response=comment_response,
+                comment_result=''
+            )
+            issue_comment = create_comment(repo_name, pr_number, comment_body)
             return
         else:
             self.log(f"account `{sender}` has permission to send commands to bot")
@@ -104,6 +119,8 @@ class EESSIBotSoftwareLayer(PyGHee):
             comment_received = request_body['comment']['body']
             self.log(f"comment action '{action}' is handled")
         else:
+            # NOTE we do not report this with a new PR comment, because otherwise any update to a comment would
+            # let the bot add a response (would be very noisy); we might add a debug mode later
             self.log(f"comment action '{action}' not handled")
             return
 
@@ -139,18 +156,11 @@ class EESSIBotSoftwareLayer(PyGHee):
         else:
             self.log(f"comment response: '{comment_response}'")
 
-        # obtain app name and format for reporting about received & processed
-        # commands
-        app_name = self.cfg[GITHUB][APP_NAME]
-        command_response_fmt = self.cfg[BOT_CONTROL][COMMAND_RESPONSE_FMT]
-
         if not any(map(get_bot_command, comment_response.split('\n'))):
             # the 'not any()' ensures that the response would not be considered a bot command itself
             # ... together with checking the sender of a comment update this aims
             # at preventing the bot to enter an endless loop in commenting on its own
             # comments
-            repo_name = request_body['repository']['full_name']
-            pr_number = request_body['issue']['number']
             comment_body = command_response_fmt.format(
                 app_name=app_name,
                 comment_response=comment_response,
@@ -329,7 +339,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         else:
             request_body = event_info['raw_request_body']
             sender = request_body['sender']['login']
-            build_msg = f"\n  - account `{sender}` has no permission to submit build jobs"
+            build_msg = f"\n  - account `{sender}` has NO permission to submit build jobs"
         return build_msg
 
     def handle_bot_command_show_config(self, event_info, bot_command):
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -135,6 +135,8 @@ class EESSIBotSoftwareLayer(PyGHee):
             self.log("update to comment is empty")
             return
 
+        comment_update += f"\n<!-- ADD NEW COMMENTS BELOW THIS LINE -->\n"
+
         if not any(map(get_bot_command, comment_update.split('\n'))):
             # the 'not any()' ensures that the update would not be considered a bot command itself
             # ... together with checking the sender of a comment update this aims
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -103,13 +103,21 @@ class EESSIBotSoftwareLayer(PyGHee):
         #      run the update through the function checking for a bot command.
         if check_command_permission(sender) is False:
             self.log(f"account `{sender}` has NO permission to send commands to bot")
-            comment_response = f"\n- account `{sender}` has NO permission to send commands to the bot"
-            comment_body = command_response_fmt.format(
-                app_name=app_name,
-                comment_response=comment_response,
-                comment_result=''
-            )
-            issue_comment = create_comment(repo_name, pr_number, comment_body)
+            # need to ensure that the bot is not responding on its own comments
+            # as a quick implementation we check if the sender name contains '[bot]'
+            # FIXME improve this by querying (and caching) information about the sender of
+            #       an event ALTERNATIVELY we could postpone this test a bit until we
+            #       have parsed the comment and know if it contains any bot command
+            if not sender.endswith('[bot]'):
+                comment_response = f"\n- account `{sender}` has NO permission to send commands to the bot"
+                comment_body = command_response_fmt.format(
+                    app_name=app_name,
+                    comment_response=comment_response,
+                    comment_result=''
+                )
+                issue_comment = create_comment(repo_name, pr_number, comment_body)
+            else:
+                self.log(f"account `{sender}` seems to be a bot instance itself, hence not creating a new PR comment")
             return
         else:
             self.log(f"account `{sender}` has permission to send commands to bot")
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -81,9 +81,11 @@ class EESSIBotSoftwareLayer(PyGHee):
             self.log(f"comment created: '{comment_diff}'")
         elif action == 'edited':
             comment_old = request_body['changes']['body']['from']
+            self.log(f"comment edited: OLD '{comment_old}'")
             comment_new = request_body['comment']['body']
+            self.log(f"comment edited: NEW '{comment_new}'")
             comment_diff = comment_new.replace(comment_old, '')
-            self.log(f"comment edited: '{comment_diff}'")
+            self.log(f"comment edited: DIFF '{comment_diff}'")
         comment_update = ''
         for line in comment_diff.split('\n'):
             self.log(f"searching line '{line}' for bot command")
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -223,7 +223,8 @@ class EESSIBotSoftwareLayer(PyGHee):
         gh = github.get_instance()
         repo = gh.get_repo(repo_name)
         pull_request = repo.get_pull(pr.number)
-        pull_request.create_issue_comment(comment)
+        comment = pull_request.create_issue_comment(comment)
+        return comment.id
 
     def handle_pull_request_event(self, event_info, log_file=None):
         """
@@ -267,7 +268,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         help_msg += "\n  - Commands must be sent with a **new** comment (edits of existing comments are ignored)."
         help_msg += "\n  - A comment may contain multiple commands, one per line."
         help_msg += "\n  - Every command begins at the start of a line and has the syntax `bot: COMMAND [ARGUMENTS]*`"
-        help_msg += "\n  - Currently supported COMMANDs are: `help`"
+        help_msg += "\n  - Currently supported COMMANDs are: `help`, `build`, `showconfig`"
         return help_msg
 
     def handle_bot_command_build(self, event_info, bot_command):
@@ -301,8 +302,14 @@ class EESSIBotSoftwareLayer(PyGHee):
 #    def handle_bot_command_status(self, event_info, bot_command):
 #        return
 
-#    def handle_bot_command_show_config(self, event_info, bot_command):
-#        return
+    def handle_bot_command_showconfig(self, event_info, bot_command):
+        self.log("processing bot command 'showconfig'")
+        gh = github.get_instance()
+        repo_name = event_info['raw_request_body']['repository']['full_name']
+        pr_number = event_info['raw_request_body']['pull_request']['number']
+        pr = gh.get_repo(repo_name).get_pull(pr_number)
+        comment_id = self.handle_pull_request_opened_event(event_info, pr)
+        return f"added comment #{comment_id} to show configuration"
 
     def start(self, app, port=3000):
         """starts the app and log information in the log file
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -89,10 +89,10 @@ class EESSIBotSoftwareLayer(PyGHee):
             self.log(f"searching line '{line}' for bot command")
             match = re.search('^bot: (.*)$', line)
             if match:
-                self.log(f"found bot command: {match.group(1)}")
+                self.log(f"found bot command: '{match.group(1)}'")
                 comment_update += "<hr/>\n received bot command "
                 comment_update += f"{match.group(1)}\n"
-        self.log(f"comment update: {comment_update}")
+        self.log(f"comment update: '{comment_update}'")
         if len(comment_update):
             repo_name = request_body['repository']['full_name']
             pr_number = int(request_body['issue']['number'])
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -127,9 +127,9 @@ class EESSIBotSoftwareLayer(PyGHee):
                 try:
                     ebc = EESSIBotCommand(bot_command)
                 except EESSIBotCommandError as bce:
-                    self.log(f"ERROR: parsing {bot_command=} failed with {bce.args}")
-                    comment_update += f"\n- parsing `{bot_command=}` received"
-                    comment_update += f" from `{sender=}` failed"
+                    self.log(f"ERROR: parsing bot command '{bot_command}' failed with {bce.args}")
+                    comment_update += f"\n- parsing bot command `{bot_command}` received"
+                    comment_update += f" from sender `{sender}` failed"
                     continue
                 commands.append(ebc)
                 self.log(f"found bot command: '{bot_command}'")
@@ -171,7 +171,7 @@ class EESSIBotSoftwareLayer(PyGHee):
                 update_pr_comment(event_info, comment_update)
                 continue
             except Exception as err:
-                log(f"Unexpected {err=}, {type(err)=}")
+                log(f"Unexpected err={err}, type(err)={type(err)}")
                 raise
 
         self.log("issue_comment event handled!")
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -94,8 +94,8 @@ class EESSIBotSoftwareLayer(PyGHee):
             match = re.search('^bot: (.*)$', line)
             if match:
                 self.log(f"found bot command: '{match.group(1)}'")
-                comment_update += "\n<hr/>\n- received bot command "
-                comment_update += f"`{match.group(1)}`\n"
+                comment_update += "\n- received bot command "
+                comment_update += f"`{match.group(1)}`"
         self.log(f"comment update: '{comment_update}'")
         if len(comment_update):
             repo_name = request_body['repository']['full_name']
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -275,7 +275,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         gh = github.get_instance()
         self.log("repository: '%s'", event_info['raw_request_body']['repository']['full_name'])
         repo_name = event_info['raw_request_body']['repository']['full_name']
-        pr_number = event_info['raw_request_body']['pull_request']['number']
+        pr_number = event_info['raw_request_body']['issue']['number']
         pr = gh.get_repo(repo_name).get_pull(pr_number)
         build_msg = ''
         if check_build_permission(pr, event_info):
@@ -306,7 +306,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         self.log("processing bot command 'showconfig'")
         gh = github.get_instance()
         repo_name = event_info['raw_request_body']['repository']['full_name']
-        pr_number = event_info['raw_request_body']['pull_request']['number']
+        pr_number = event_info['raw_request_body']['issue']['number']
         pr = gh.get_repo(repo_name).get_pull(pr_number)
         comment_id = self.handle_pull_request_opened_event(event_info, pr)
         return f"added comment #{comment_id} to show configuration"
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -13,7 +13,6 @@
 #
 # license: GPLv2
 #
-import re
 import waitress
 import sys
 
@@ -21,6 +20,7 @@ from connections import github
 from tools import config
 from tools.args import event_handler_parse
 from tools.commands import get_bot_command, EESSIBotCommand, EESSIBotCommandError
+from tools.filter import EESSIBotActionFilter
 from tools.permissions import check_command_permission
 from tools.pr_comments import update_pr_comment
 from tasks.build import check_build_permission, submit_build_jobs, get_repo_cfg
@@ -68,7 +68,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         txt = request_body['comment']['body']
         self.log(f"Comment in {issue_url} (owned by @{owner}) {action} by @{sender}: {txt}")
         # check if addition to comment includes a command for the bot, e.g.,
-        #   bot: rebuild [arch:intel] [instance:AWS]
+        #   bot: build [arch:intel] [instance:AWS]
         #   bot: cancel [job:jobid]
         #   bot: disable [arch:generic]
         # actions: created, edited
@@ -155,7 +155,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         # process commands
         for cmd in commands:
             try:
-                update = self.handle_bot_command(cmd, event_info)
+                update = self.handle_bot_command(event_info, cmd)
                 comment_update += f"\n- handling `{cmd.command}` resulted in:"
                 comment_update += update
                 update_pr_comment(event_info, comment_update)
@@ -193,7 +193,8 @@ class EESSIBotSoftwareLayer(PyGHee):
         if label == "bot:build":
             # run function to build software stack
             if check_build_permission(pr, event_info):
-                submit_build_jobs(pr, event_info)
+                # use an empty filter
+                submit_build_jobs(pr, event_info, EESSIBotActionFilter(""))
 
         elif label == "bot:deploy":
             # run function to deploy built artefacts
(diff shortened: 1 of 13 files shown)


trz42:
> @trz42 Can you look into fixing the merge conflicts?

Done.

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+        else:
+            self.log(f"comment action '{action}' not handled")
+            return
+
+        # search for commands in what is new in comment
+        # init comment_update with an empty string or later split would fail if
+        # it is None
+        comment_update = ''
+        commands = []
+        for line in comment_new.split('\n'):
```
boegel, line 125: Can skip empty lines already here:

```python
        for line in [x for x in [y.strip() in comment_new.split('\n')] if x]:

File: eessi_bot_event_handler.py [hunk 2]
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            # next line commented out to write less to log files
+            # self.log(f"searching line '{line}' for bot command")
+            line_stripped = line.strip()
+            if line.isspace() or len(line_stripped) == 0:
```
boegel, line 129: This can be simplified quite a bit (if stripped line is empty, then there's no need to check with `isspace` too):

```python
if not line_stripped:

File: eessi_bot_event_handler.py [hunk 3]
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+                # next line commented out to write less to log files
+                # self.log(f"line '{line}' is empty")
```
boegel, line 131: add `FIXME DEBUG` to comment

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            if bot_command:
+                try:
+                    ebc = EESSIBotCommand(bot_command)
+                except EESSIBotCommandError as bce:
+                    self.log(f"ERROR: parsing bot command '{bot_command}' failed with {bce.args}")
+                    # next two lines commented out to make the bot less noisy,
+                    # could be enabled again if bot commands accept arg --verbose
+                    # or bot instance has a similar setting
+                    # comment_update += f"\n- parsing bot command `{bot_command}` received"
+                    # comment_update += f" from sender `{sender}` failed"
```
boegel, line 143: Hmm, I would always do this, mistakes shouldn't be silently ignored?

Maybe this should result in the bot adding an extra comment rather that updating the comment in which the command was found, which is more visible and will trigger a notification to the account that created the comment?

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+                # or bot instance has a similar setting
+                # comment_update += f"\n- line <code>{line}</code> is not considered to contain a bot command"
+                # comment_update += "\n- bot commands begin with `bot: `, make sure"
+                # comment_update += "\n  there is no whitespace at the beginning of a line"
+
+        if comment_update == '':
+            # no update to be added, just log and return
+            self.log("update to comment is empty")
+            return
+        else:
```
boegel, line 164: We could prepend the comment here with `"Updates by bot:\n"`, so the `if` check can keep checking for empty string.

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            self.log(f"comment update: '{comment_update}'")
+
+        if not any(map(get_bot_command, comment_update.split('\n'))):
+            # the 'not any()' ensures that the update would not be considered a bot command itself
+            # ... together with checking the sender of a comment update this aims
+            # at preventing the bot to enter an endless loop in commenting on its own
+            # comments
+            update_pr_comment(event_info, comment_update)
+        else:
+            self.log(f"update '{comment_update}' is considered to contain bot command ... not updating PR comment")
```
boegel, line 174: This should be an error, no?

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+
+        # process commands
+        for cmd in commands:
```
boegel, line 177: The implementation of `handle_issue_comment_event` is getting quite long.

How about introducing two methods there are called in `handle_issue_comment_event`. Something like:

```python
commands = self.extract_bot_commands(...)
self.log("Found %d bot commands in comment ..." % len(commands))
self.process_bot_commands(commands)

File: eessi_bot_event_handler.py [hunk 8]
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            try:
+                update = self.handle_bot_command(event_info, cmd)
+                # the update to the PR comment is only done if the command is
+                # 'help'; otherwise there was too much noise and also updates
+                # were a bit confusing when different instances overwrote
+                # comments of the other instances
+                if cmd.command == 'help':
+                    comment_update += f"\n- handling `{cmd.command}` resulted in: "
+                    comment_update += update
+                    update_pr_comment(event_info, comment_update)
```
boegel, line 187: Can't we let the bot reply with a new comment instead of updating the existing comment?
Then there's no problem with multiple instances...

Ideally the reply to `help` is just a pointer to the documentation BTW ;)

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+
+                self.log(f"handling '{cmd.command}' resulted in '{update}'")
+
+            except EESSIBotCommandError as bce:
+                self.log(f"ERROR: handling {cmd.command} failed with {bce.args}")
+                # next two lines commented out to make the bot less noisy,
+                # could be enabled again if bot commands accept arg --verbose
+                # or bot instance has a similar setting
+                # comment_update += f"\n- handling `{cmd.command}` failed with {bce.args}"
+                # update_pr_comment(event_info, comment_update)
```
boegel, line 197: I would post a new comment if an error occurred, these should be visible, otherwise people will get confused that bot is not responding?

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+                continue
+            except Exception as err:
+                log(f"Unexpected err={err}, type(err)={type(err)}")
+                raise
+
         self.log("issue_comment event handled!")
```
boegel, line 203: let's also mention issue ID or URL in this log message?

boegel (changes requested):
first part of review, more coming soon...

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+#        return
+
+#    def handle_bot_command_disable(self, event_info, bot_command):
+#        return
+
+#    def handle_bot_command_cancel(self, event_info, bot_command):
+#        return
+
+#    def handle_bot_command_status(self, event_info, bot_command):
+#        return
```
boegel, line 342: I would remove this until we have these commands actually implemented?

`handle_bot_command_build` can serve as an example

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+        handler_name = f"handle_bot_command_{cmd}"
+        if hasattr(self, handler_name):
+            handler = getattr(self, handler_name)
+            self.log(f"Handling bot command {cmd}")
+            return handler(event_info, bot_command)
+        else:
+            self.log(f"No handler for command '{cmd}'")
+            raise EESSIBotCommandError(f"unknown command '{cmd}'; use `bot: help` for usage information")
+
+    def handle_bot_command_help(self, event_info, bot_command):
```
boegel, line 305: missing docstring?

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+        help_msg = "\n  **How to send commands to bot instances**"
+        help_msg += "\n  - Commands must be sent with a **new** comment (edits of existing comments are ignored)."
+        help_msg += "\n  - A comment may contain multiple commands, one per line."
+        help_msg += "\n  - Every command begins at the start of a line and has the syntax `bot: COMMAND [ARGUMENTS]*`"
+        help_msg += "\n  - Currently supported COMMANDs are: `help`, `build`, `showconfig`"
+        return help_msg
+
+    def handle_bot_command_build(self, event_info, bot_command):
```
boegel, line 313: missing docstring?

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+        gh = github.get_instance()
```
boegel, line 314: Should there be a `self.gh` so we don't need to create a new instance every time?

Not sure that's wise actually, since it'll be long-lived, just a random thought

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+#    def handle_bot_command_disable(self, event_info, bot_command):
+#        return
+
+#    def handle_bot_command_cancel(self, event_info, bot_command):
+#        return
+
+#    def handle_bot_command_status(self, event_info, bot_command):
+#        return
+
+    def handle_bot_command_showconfig(self, event_info, bot_command):
```
boegel, line 344: Nitpicking: maybe rename `showconfig` to `show_config`?

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+        self.log("processing bot command 'showconfig'")
+        gh = github.get_instance()
+        repo_name = event_info['raw_request_body']['repository']['full_name']
+        pr_number = event_info['raw_request_body']['issue']['number']
+        pr = gh.get_repo(repo_name).get_pull(pr_number)
```
boegel, line 349: Wouldn't it make more sense to move this stuff to `handle_bot_command`, and pass down `pr` into `handle_bot_command_*`?

eessi_bot_event_handler.py
```diff
@@ -21,13 +21,22 @@
...
+from tools.pr_comments import update_pr_comment
+from tasks.build import check_build_permission, submit_build_jobs, get_repo_cfg, get_architecturetargets
 from tasks.deploy import deploy_built_artefacts
 
 from pyghee.lib import PyGHee, create_app, get_event_info, read_event_from_json
 from pyghee.utils import log
 
 
+GITHUB = "github"
+APP_NAME = "app_name"
```
boegel, line 36: These are also defined in `task/build.py`, which suggests there should be a common `constants` module to import these from?

Can be done in a follow-up PR of course

tasks/build.py
```diff
@@ -67,96 +69,106 @@
...
     """
-    cfg = config.read_config()
+    fn = sys._getframe().f_code.co_name
+
+    # cfg = config.read_config()
     buildenv = cfg[BUILDENV]
 
     jobs_base_dir = buildenv.get(JOBS_BASE_DIR)
-    log("jobs_base_dir '%s'" % jobs_base_dir)
+    log(f"{fn}(): jobs_base_dir '{jobs_base_dir}'")
```
boegel, line 93: Do we already have an open issue to enhance the `log` function so it automatically adds the function from which the log message was sent?

That would be way cleaner than spreading `{fn}()` all over the place in `log` calls, and shouldn't be very difficult I think.

No blocker here, can be cleaned up later since we're already doing this all over the place

tasks/build.py
```diff
@@ -67,96 +69,106 @@
...
     container_cachedir = buildenv.get(CONTAINER_CACHEDIR)
-    log("container_cachedir '%s'" % container_cachedir)
+    log(f"{fn}(): container_cachedir '{container_cachedir}'")
     config_data[CONTAINER_CACHEDIR] = container_cachedir
 
     cvmfs_customizations = {}
     try:
         cvmfs_customizations_str = buildenv.get(CVMFS_CUSTOMIZATIONS)
-        log("cvmfs_customizations '%s'" % cvmfs_customizations_str)
+        log("cvmfs_customizations '{cvmfs_customizations_str}'")
```
boegel, line 120: `{fn}()` is missing ;)

tasks/build.py
```diff
@@ -271,9 +292,11 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
     Args:
         repo_name (string): pr base repo name
         branch_name (string): pr branch name
-        pr (object): pr details
+        pr (github.PullRequest.Pullrequest): object to interact with pull request
         arch_job_dir (string): location of arch_job_dir
     """
+    # fn = sys._getframe().f_code.co_name
```
boegel, line 298: These should be cleaned up too when `log` does this automagically

tasks/build.py
```diff
@@ -324,28 +349,40 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
     Args:
-        pr: github.PullRequest.Pullrequest object
-        event_dir: base directory for all jobs created for this event
-        build_env_cfg: build env configuration
-        arch_map: maps OS/SW_DIR to Slurm parameters
-        repocfg: repository config
+        pr (github.PullRequest.Pullrequest): object to interact with pull request
+        cfg (dict): dictionary holding full configuration (by default defined in app.cfg)
+        event_info (dict): event received by event_handler
+        action_filter (EESSIBotActionFilter): used to filter which jobs shall be prepared
```
boegel, line 360:
```suggestion
        action_filter (EESSIBotActionFilter instance): used to filter which jobs shall be prepared
```

tasks/build.py
```diff
@@ -324,28 +349,40 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
     log(f"{fn}(): pr.base.repo.full_name '{base_repo_name}'")
 
     base_branch_name = pr.base.ref
     log(f"{fn}(): pr.base.repo.ref '{base_branch_name}'")
 
+    # create run dir (base directory for potentially several jobs)
+    # TODO may still be too early (before we get to any actual job being
+    #      prepared below when calling 'download_pr')
+    # instead of using a run_dir, maybe just create a unique dir for each
+    # job to be submitted? thus we could easily postpone the create_pr_dir
```
boegel, line 382: open issue for this TODO?

tasks/build.py
```diff
@@ -359,7 +396,20 @@ def prepare_jobs(pr, event_dir, build_env_cfg, arch_map, repocfg):
             if repo_id != "EESSI-pilot" and repo_id not in repocfg:
                 log(f"{fn}(): skipping repo {repo_id}, it is not defined in repo config {repocfg[REPOS_CFG_DIR]}")
                 continue
-            job_dir = os.path.join(event_dir, arch_dir, repo_id)
+
+            # if filter exists, check filter against context = (arch, repo, app_name)
+            #   true --> log & go on
```
boegel, line 401: `log & go on in this iteration of for loop`

tasks/build.py
```diff
@@ -359,7 +396,20 @@ def prepare_jobs(pr, event_dir, build_env_cfg, arch_map, repocfg):
...
+            #   false --> log & continue
```
boegel, line 402: `log & continue` to next iteration of `for` loop

tasks/build.py
```diff
@@ -324,28 +349,40 @@ def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
...
     base_branch_name = pr.base.ref
     log(f"{fn}(): pr.base.repo.ref '{base_branch_name}'")
 
+    # create run dir (base directory for potentially several jobs)
+    # TODO may still be too early (before we get to any actual job being
+    #      prepared below when calling 'download_pr')
+    # instead of using a run_dir, maybe just create a unique dir for each
+    # job to be submitted? thus we could easily postpone the create_pr_dir
+    # call to just before download_pr
+    ym, pr_id, run_dir = create_pr_dir(pr, cfg, event_info)
```
boegel, line 384: no idea what `ym` is, deserves better variable name

tasks/build.py
```diff
@@ -507,49 +558,50 @@ def submit_job(job, submitted_jobs, build_env_cfg, ym, pr_id):
...
-    log(f'submit_build_jobs(): sbatch err: {cmdline_error}')
+    log(f"{fn}(): sbatch out: {cmdline_output}")
+    log(f"{fn}(): sbatch err: {cmdline_error}")
 
     job_id = cmdline_output.split()[3]
-    submitted_jobs.append(job_id)
-    symlink = os.path.join(build_env_cfg[JOBS_BASE_DIR], ym, pr_id, job_id)
-    log(f"jobs_base_dir: {build_env_cfg[JOBS_BASE_DIR]}, ym: {ym}, pr_id: {pr_id}, job_id: {job_id}")
+    ym = job.year_month
+    pr_id = job.pr_id
```
boegel, line 566: This saves 3 characters, seems overkill, just use `job.pr_id`?

Also use `job.year_month` rather than the cryptic `ym`

tasks/build.py
```diff
@@ -587,60 +640,60 @@ def create_pr_comment(job, job_id, app_name, pr_number, repo_name, gh, symlink):
...
 
 
-def submit_build_jobs(pr, event_info):
+def submit_build_jobs(pr, event_info, action_filter):
     """Build from the pr by fetching data for build environment cofinguration, downloading pr,
        running jobs and adding comments
 
     Args:
-        pr (object): _description_
+        pr (github.PullRequest.Pullrequest): object to interact with pull request
```
boegel, line 648: instance

tasks/build.py
```diff
@@ -587,60 +640,60 @@ def create_pr_comment(job, job_id, app_name, pr_number, repo_name, gh, symlink):
...
         event_info (string): event received by event_handler
+        action_filter (EESSIBotActionFilter): used to filter which jobs shall be prepared
```
boegel, line 650: instance

tasks/build.py
```diff
@@ -587,60 +640,60 @@ def create_pr_comment(job, job_id, app_name, pr_number, repo_name, gh, symlink):
...
-    # [architecturetargets]
-    arch_target_map = get_architecturetargets()
+    # return if no jobs to be submitted
+    if not jobs:
+        log(f"{fn}(): no jobs ({len(jobs)}) to be submitted")
+        return "no jobs were prepared"
 
-    # [directory structure]
-    ym, pr_id, run_dir = create_pr_dir(pr, build_env_cfg[JOBS_BASE_DIR], event_info)
+    # obtain handle to GH
```
boegel, line 666: `GH` -> `GitHub`

tasks/build.py
```diff
@@ -587,60 +640,60 @@ def create_pr_comment(job, job_id, app_name, pr_number, repo_name, gh, symlink):
...
-        pr_comment_id = create_pr_comment(job, job_id, app_name, int(pr.number), repo_name, gh, symlink)
+        # report submitted job
+        pr_comment_id = create_pr_comment(job, job_id, app_name, pr, gh, symlink)
 
         # create _bot_job<jobid>.metadata file in submission directory
-        create_metadata_file(job, job_id, repo_name, int(pr.number), pr_comment_id)
+        create_metadata_file(job, job_id, pr, pr_comment_id)
+
+    return_msg = f"created jobs: {', '.join(job_ids)}"
+    return return_msg
```
boegel, line 683:
```python
    return f"created jobs: {', '.join(job_ids)}"

File: tools/commands.py [hunk 31]
```diff
@@ -0,0 +1,62 @@
...
+        return match.group(1).rstrip()
+    else:
+        return None
+
+
+class EESSIBotCommandError(Exception):
+    pass
+
+
+class EESSIBotCommand:
```
boegel, line 41: missing docstring?

tools/commands.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, cmd_str):
+        cmd_as_list = cmd_str.split()
+        self.command = cmd_as_list[0]
+        if len(cmd_as_list) > 1:
+            arg_str = " ".join(cmd_as_list[1:])
+            try:
+                self.action_filters = EESSIBotActionFilter(arg_str)
+            except EESSIBotActionFilterError as baf:
```
boegel, line 49: `baf` -> `err`?

tools/commands.py
```diff
@@ -0,0 +1,62 @@
...
+                reason = baf.args
+                log(f"ERROR: EESSIBotActionFilterError - {reason}")
```
boegel, line 51: This works fine I think?

```suggestion
                log(f"ERROR: EESSIBotActionFilterError - {err.args}")

File: tools/filter.py [hunk 34]
```diff
@@ -0,0 +1,150 @@
...
+        for _filter in filter_string.split():
+            try:
+                self.add_filter_from_string(_filter)
+            except EESSIBotActionFilterError:
+                raise
+            except Exception as err:
+                log(f"Unexpected err={err}, type(err)={type(err)}")
+                raise
+
+    def clear_all(self):
```
boegel, line 44: Missing docstring?

tools/filter.py
```diff
@@ -0,0 +1,150 @@
...
+        Adds a filter
+
+        Args:
+            component (string): any prefix of 'architecture', 'instance', 'job' or 'repository'
+            pattern (string): regex that is applied to a string representing the component
+        """
+        # check if component is supported
+        full_component = None
+        for cis in FILTER_COMPONENTS:
+            if cis.startswith(component):
```
boegel, line 58: Right now we don't have filter components that partially overlap, but what if we introduce one?

At the very least we should add a comment above definition of `FILTER_COMPONENTS` to warn about this, and maybe this code needs to be made a bit more intelligent too (collect all prefix matches, only conclude we found match if there's only one match)

tools/filter.py
```diff
@@ -0,0 +1,150 @@
...
+                full_component = cis
+                break
+        if full_component:
+            log(f"processing component {component}")
+            # if full_component == architecture replace - with / in pattern
+            if full_component == 'architecture':
```
boegel, line 64: `'architecture'` should be a constant, so above:

```python
FILTER_COMPONENT_ARCH = 'architecture'
FILTER_COMPONENTS = [FILTER_COMPONENT_ARCH, ...]

File: tools/filter.py [hunk 37]
```diff
@@ -0,0 +1,150 @@
...
+        # examples:
+        #   arch:intel instance:AWS --> rebuild for all repos for intel architectures on AWS
+        #   arch:generic --> rebuild for all repos for generic architectures anywhere
+        #   repository:nessi.no-2022.11 --> disable all builds for nessi.no-2022.11
+        # we iterate over all defined filters
+        for af in self.action_filters:
+            if af.component in context:
+                value = context[af.component]
+                # replace - with / in architecture component
+                if af.component == 'architecture':
```
boegel, line 142: use constant instead of `'architecture'`

tools/permissions.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pyghee.utils import log
+from tools import config
+
+BOT_CONTROL = "bot_control"
+COMMAND_PERMISSION = "command_permission"
+
+
+def check_command_permission(account):
+    """check if the GH account is authorized to send commands to the bot
```
boegel, line 21: `GH` -> `GitHub`

tools/permissions.py
```diff
@@ -0,0 +1,46 @@
...
+import sys
+
+from pyghee.utils import log
+from tools import config
+
+BOT_CONTROL = "bot_control"
+COMMAND_PERMISSION = "command_permission"
+
+
+def check_command_permission(account):
```
boegel, line 20: Checking of `build` and `deploy` permission should also come from here (eventually)?

boegel (changes requested):
Minor issue: tests are missing for stuff in `tools/commands.py` for example, but that's not a huge disaster (not worth blocking PR over imho)

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
         """
         request_body = event_info['raw_request_body']
         issue_url = request_body['issue']['url']
-        comment_author = request_body['comment']['user']['login']
-        comment_txt = request_body['comment']['body']
-        self.log("Comment posted in %s by @%s: %s", issue_url, comment_author, comment_txt)
+        action = request_body['action']
+        sender = request_body['sender']['login']
+        owner = request_body['comment']['user']['login']
+        # next two lines are commented out and replaced to write less to log files
```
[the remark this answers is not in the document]
trz42: Maybe this one https://github.com/EESSI/eessi-bot-software-layer/issues/91

eessi_bot_event_handler.py
boegel: Could be problematic if the token expires, and `github.*` caches the instance already anyway, so nevermind

boegel (changes requested):

eessi_bot_event_handler.py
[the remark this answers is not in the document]
trz42: Originally the full body of the event was written to the log. This was a bit too much _or_ is too much for non-development work. I'll remove the commented lines.

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+        #   1. check permission
+        #   2. skip any comment updates that were done by the bot itself --> we
+        #      prevent the bot entering an endless loop where it reacts on
+        #      updates to comments it made itself
+        #      NOTE this assumes that the sender of the event is corresponding to
+        #      the bot if the bot updates comments itself and that the bot is not
+        #      given permission in the configuration setting 'command_permission'
+        #      ... in order to prevent surprises we should be careful what the bot
+        #      adds to comments, for example, before updating a comment it could
+        #      run the update through the function checking for a bot command.
```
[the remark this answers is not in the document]
trz42: That's right, it should be fine. It might be further improved by checking that the bot's name is not listed in the setting `command_permission`.

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            return
+
+        # search for commands in what is new in comment
+        # init comment_update with an empty string or later split would fail if
+        # it is None
+        comment_update = ''
+        commands = []
+        for line in comment_new.split('\n'):
+            # next line commented out to write less to log files
+            # self.log(f"searching line '{line}' for bot command")
```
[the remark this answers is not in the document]
trz42: Hmm, I'd rather remove it. It was useful while developing.

eessi_bot_event_handler.py
trz42: Hmm, this doesn't look complete to me, maybe rather
```python
for line in [x for x in [y.strip() for y in comment_new.split('\n')] if x]:
```
I'll give it a try. Will add a short comment so it will be understandable later too.

eessi_bot_event_handler.py
trz42: I rather removed such comments. The logging was mainly intended to help during development.

eessi_bot_event_handler.py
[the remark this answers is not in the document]
trz42: Hmm, rather replaced the two lines with a FIXME that we may add the body of the event to the log message if debug log level is set.

eessi_bot_event_handler.py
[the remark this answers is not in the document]
trz42: The bot's _name_ should not be confused with the `app_name` setting in `app.cfg`. We need to figure out how the bot's name can be figured out.

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            # self.log(f"comment created:\n########\n{comment_new.rstrip()}\n########")
+            self.log(f"comment action '{action}' is handled")
+        else:
+            self.log(f"comment action '{action}' not handled")
+            return
+
+        # search for commands in what is new in comment
+        # init comment_update with an empty string or later split would fail if
+        # it is None
+        comment_update = ''
```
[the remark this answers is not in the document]
trz42: Should be changed to something like

<details><summary>Updates by the bot instance <code>AWS</code>
<em>(click for details)</em></summary>

- parsing the bot command `bot: install TensorFlow`, received from sender `wizard`, failed
_or_
- received bot command `bot: build i:AWS repo:eessi-2023.04 arc:aarch64/graviton2` from sender `:D`
  - expanded format: `bot: build instance:AWS ...`
- handling `bot: build instance:AWS ...` resulted in: submitted job `12456`
_or_
- handling `bot: build arch:riscv/generic` failed with "no jobs submitted"
</details>

eessi_bot_event_handler.py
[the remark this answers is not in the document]
trz42: Replaced commented lines with a FIXME.

eessi_bot_event_handler.py
trz42: Will try
```python
        for line in [x for x in [y.strip() for y in comment_received.split('\n')] if x]:$
```

eessi_bot_event_handler.py
trz42: Replaced commented lines with a FIXME

eessi_bot_event_handler.py
trz42: Ack. Keeping that type of information in the response to a bot command.

eessi_bot_event_handler.py
```diff
@@ -60,9 +69,137 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+                    # or bot instance has a similar setting
+                    # comment_update += f"\n- parsing bot command `{bot_command}` received"
+                    # comment_update += f" from sender `{sender}` failed"
+                    continue
+                commands.append(ebc)
+                self.log(f"found bot command: '{bot_command}'")
+                comment_update += "\n- received bot command "
+                comment_update += f"`{bot_command}`"
+                comment_update += f" from `{sender}` (expanded command format: `{ebc.to_string()}`)"
+            # next two lines are commented out to write less to log files
```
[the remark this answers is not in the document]
trz42: Replaced commented lines with a FIXME.

eessi_bot_event_handler.py
trz42: This is kept. The additional text is implemented via a configuration setting.

eessi_bot_event_handler.py
trz42: Hmm, so you're thinking of updating the PR comment with something like "Oops response message seems to contain a bot command. It is not displayed here to prevent the bot from entering an endless loop of commands. Please, check the logs at the bot instance for more information."?

I'll create a FIXME.

eessi_bot_event_handler.py
trz42: Hmm, indeed.

However, the data handed over from extracting commands to processing them is not just a list of `commands`. These functions would at least need `event_info`, the comment GitHub instance (or we need to query GitHub a second time), plus possibly other state. If we move them to a separate module, we also need to take of config (now accessible as `self.cfg`), logging is affected ...

I agree it's long. Tend to leave it as is and rather open an issue for refactoring this later.

eessi_bot_event_handler.py
trz42: Yes. It creates new comments now.

Yeah, help is just a toy to play with commands.

eessi_bot_event_handler.py
trz42: Yes. This is re-implemented now.

eessi_bot_event_handler.py
trz42: True. Added the `request_body['issue']['url']`.

eessi_bot_event_handler.py
```diff
@@ -138,6 +284,72 @@ def handle_pull_request_event(self, event_info, log_file=None):
...
+        """
+        cmd = bot_command.command
+        handler_name = f"handle_bot_command_{cmd}"
+        if hasattr(self, handler_name):
+            handler = getattr(self, handler_name)
+            self.log(f"Handling bot command {cmd}")
+            return handler(event_info, bot_command)
+        else:
+            self.log(f"No handler for command '{cmd}'")
+            raise EESSIBotCommandError(f"unknown command '{cmd}'; use `bot: help` for usage information")
```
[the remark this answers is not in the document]
trz42: OK, will create an issue for that.

eessi_bot_event_handler.py
trz42: done

eessi_bot_event_handler.py
trz42: In addition to `event_info` and `bot_command`? Maybe. Every method needs something slightly different. `_help` doesn't need any additional data, `_build` needs the `event_info` (for passing it down to existing functions) and `bot_command`. `_show_config` needs `event_info`. Both `_build` and `_show_config` use `event_info` to init an object for the pull request and pass that down to subsequent function calls. An improvement could be to put duplicate code into a function `get_pr_from_issue_event(event_info)` that returns the pull request object (we have to be a bit careful because different events have different data ... or at least name attributes differently).

tasks/build.py
trz42: Hafsa was working on something like this (issue was https://github.com/EESSI/eessi-bot-software-layer/issues/110). However, it is not generally applicable. The logging definitely needs more thinking/rework.

tasks/build.py
trz42: added

tasks/build.py
```diff
@@ -67,96 +69,106 @@
...
     load_modules = buildenv.get(LOAD_MODULES, None)
-    log("load_modules '%s'" % load_modules)
+    log(f"{fn}(): load_modules '{load_modules}'")
     config_data[LOAD_MODULES] = load_modules
 
     return config_data
 
 
-def get_architecturetargets():
+def get_architecturetargets(cfg):
```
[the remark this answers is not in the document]
trz42: yep

tasks/build.py
trz42: Replaced everywhere with `year_month`

tasks/build.py
trz42: done

tasks/build.py
trz42: done

tasks/build.py
trz42: changed

tasks/build.py
trz42: sure

tasks/build.py
trz42: yep

tasks/build.py
trz42: :+1:

tasks/build.py
trz42: done

tools/commands.py
trz42: added

tools/commands.py
trz42: changed

tools/commands.py
trz42: sure

tools/filter.py
trz42: added

tools/filter.py
trz42: True. I can't think of two components with the same prefix ... Jokes aside, I added a note at the definition of `FILTER_COMPONENTS` and also to `add_filter` (that it assumes that no two filter share the same prefix).

tools/filter.py
trz42: ack

tools/filter.py
trz42: done

tools/permissions.py
trz42: yes makes sense

tools/permissions.py
trz42: ✅

tools/pr_comments.py
```diff
@@ -81,15 +95,22 @@ def update_comment(cmnt_id, pr, update, log_file=None):
             log_file=log_file)
 
 
-def create_comment(repo_name, pr_number, comment):
-    """create a comment on a pr
+def update_pr_comment(event_info, update):
+    """
+    Updates a comment determined from an event.
```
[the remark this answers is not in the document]
trz42: yep

eessi_bot_event_handler.py
boegel: OK, let's open the issue.
I don't think we should move it to a separate module, it can be methods which can access `self.cfg`, but if passing around too much data is required, then maybe it doesn't make much sense.

tools/filter.py
boegel: OK, good enough for now

eessi_bot_event_handler.py
```diff
@@ -60,10 +71,151 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+        #   2. skip any comment updates that were done by the bot itself
+        #      --> thus we prevent the bot from entering an endless loop
+        #          where it reacts on updates to comments it made itself
+        #      NOTE this assumes that the sender of the event is corresponding to
+        #      the bot if the bot updates comments itself and that the bot is not
+        #      given permission in the configuration setting 'command_permission'
+        #      ... in order to prevent surprises we should be careful what the bot
+        #      adds to comments, for example, before updating a comment it could
+        #      run the update through the function checking for a bot command.
+        if check_command_permission(sender) is False:
```
boegel, line 104: Since this is triggered for *every* issue comment being made, it's done too early, since for accounts that don't have permission to send commands to the bot this will immediately trigger a `account XXX has NO permission to send commands to the bot` (see https://github.com/boegel/software-layer/pull/24#issuecomment-1576845077).

That should only be done if the comment effectively includes bot commands.

Let's try and make this the first thing we fix after merging this PR...

PR #177, opened 2023-04-14
some fixes for letting the job manager handle non bot jobs correctly 

trz42:
- need to remove non bot jobs earlier, because
  - `process_running_jobs()` does not double-check if a job is a bot's job or not
  - current jobs are determined via `squeue` so could include non bot jobs (again for every iteration) 
- using current jobs for processing running jobs instead of known jobs
  - a running job may not yet be in the list of known jobs (that is created at the start or in the last iteration)
  - when the current jobs are used we need to make sure it only includes bot jobs (therefore non bot jobs need to be removed from the list as early as possible)

- somewhat tested with recent jobs to build compat layer (and discovered while running interactive debug sessions)

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -676,6 +676,10 @@ def main():
             #        " %s due to parameter '--jobs %s'" % (
             #          nj,opts.jobs), job_manager.logfile)
 
+        # remove non bot jobs from current_jobs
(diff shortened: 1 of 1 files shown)


bedroge (approved):
Looks good to me, makes sense to filter out the non-bot jobs a bit earlier.

PR #178, opened 2023-04-26
replay a GitHub event locally at a bot instance

### LINKED ISSUE #94: Replay a GitHub event locally
Events are logged by the bot (via PyGHee). For debugging purposes it could be useful to replay any event locally.

jonas-lq:
Fixes #94 

PR for being able to replay events stored in the file system.

Triggered by commenting "replay <event id>" on the pr (Your GitHub app must subscribe to the event "Issue comment")

There were already some support for replaying events (see line 200 in `eessi_bot_event_handler.py`), but this is outdated and would not work even if the specified file contained both the header and body (PyGHee logs events as separate files for body and header). Should I change the code here to utilise my newly implemented function?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ appbackup.cfg
 venv_bot_p37
 *.swo
 events_log/
+event_links/
+venv
--- a/app.cfg.example
+++ b/app.cfg.example
@@ -120,6 +120,7 @@ arch_target_map = { "linux/x86_64/generic" : "--constraint shape=c4.2xlarge", "l
 [event_handler]
 # path to the log file to log messages for event handler
 log_path = /path/to/eessi_bot_event_handler.log
+event_links_path = /path/to/event_links
 
 
 [job_manager]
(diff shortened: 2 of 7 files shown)


trz42 (changes requested):
Looks already quite nice. The PR does more than what #94 asked for. #94 just suggested to have a means to replay an event **locally** (where a bot instance is running). A couple of suggestions:

- remove the ability to send the `replay` command from GitHub (it will be impractical to use or worse lead to unintended consequences)
- think of improving the efficiency, as of now `os.walk()` will traverse the whole directory of the bot instance and access 100k or even millions of entries
  - one could/should make use of the directory structure starting at `events_log`
  - one could also think of improving PyGHee such that it creates symlinks, eg, `event_id` -> directory that contains the header & body json
    - thus there need to be no lookup/traversal of the file system
- there could be a new small script, say `replay_event.py` that imports the `replay_event` method
  - this probably requires to move the core of the method `replay_event` out of the EESSIBotEventHandler class to another module, say `tools/replay_events.py`
  - thus we could have a tool that replays an event and if we later decide to use that from the bot (event handler) we can just reuse it
- a follow-up PR could improve the replay_event tool to limit the effects of replaying an event, for example, only replay it for certain architectures or job ids
  - this might use code developed for #172

trz42:
Closing this one. We may revisit it at a later time.

Issue #176, opened 2023-04-11
Add default pr comments


jonas-lq:
The pr comments defined in app.cfg.example are required by the event handler and job manager. It could be beneficial to not require them, and instead have default comments if some of them are not specified.

See: https://github.com/EESSI/eessi-bot-software-layer/pull/170#discussion_r1162399300

PR #181, opened 2023-05-12
Added default comments

jonas-lq:
Fix for #176


trz42:
This one requires updates. Closing it now, but we can revisit it later.

PR #182, opened 2023-05-13
add comment id to metadata uploaded to S3

trz42:
The comment id can be used by the ingestion script of the filesystem layer to update the PR comment corresponding to the job that has generated a tarball (software package(s), module file(s) and lmod cache update).


trz42:
Replaced by a series of smaller PRs:

- #247 
- #249 
- #250 
- #253

PR #189, opened 2023-06-23
prefer bot/check-build.sh script over bot/check-result.sh in build job script

boegel:
While writing the [documentation for the bot](https://www.eessi.io/docs/bot/), I realized that `check-result.sh` is a bit of a misnomer, since eventually we may also want a `check-test.sh` and `check-deploy.sh` script.

By still considering `check-result.sh` as a fallback when `check-build.sh` is not there, we can transition easily to `check-build.sh` without causing trouble...

### DIFF
--- a/scripts/bot-build.slurm
+++ b/scripts/bot-build.slurm
@@ -32,12 +32,20 @@ else
     exit 1
 fi
 echo "bot/build.sh finished"
-CHECK_RESULT_SCRIPT=bot/check-result.sh
-if [ -f ${CHECK_RESULT_SCRIPT} ]; then
-    echo "${CHECK_RESULT_SCRIPT} script found in '${PWD}', so running it!"
-    ${CHECK_RESULT_SCRIPT}
+CHECK_BUILD_SCRIPT=bot/check-build.sh
+
+# temporarily also still consider bot/check-result.sh script as a fallback
+CHECK_RESULT_SCRIPT=bot/check-results.sh
+if [ ! -f ${CHECK_BUILD_SCRIPT} ] && [ -f ${CHECK_RESULT_SCRIPT} ]; then
+    echo "${CHECK_BUILD_SCRIPT} not found in ${PWD}, but ${CHECK_RESULT_SCRIPT} is found, so using that as fallback..."
+    CHECK_BUILD_SCRIPT=${CHECK_RESULT_SCRIPT}
+fi
+
+if [ -f ${CHECK_BUILD_SCRIPT} ]; then
+    echo "${CHECK_BUILD_SCRIPT} script found in '${PWD}', so running it!"
+    ${CHECK_BUILD_SCRIPT}
 else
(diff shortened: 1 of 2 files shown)


scripts/bot-build.slurm
```diff
@@ -32,12 +32,20 @@ else
...
 fi
 echo "bot/build.sh finished"
-CHECK_RESULT_SCRIPT=bot/check-result.sh
-if [ -f ${CHECK_RESULT_SCRIPT} ]; then
-    echo "${CHECK_RESULT_SCRIPT} script found in '${PWD}', so running it!"
-    ${CHECK_RESULT_SCRIPT}
+CHECK_BUILD_SCRIPT=bot/check-build.sh
+
+# temporarily also still consider bot/check-result.sh script as a fallback
+CHECK_RESULT_SCRIPT=bot/check-results.sh
```
trz42, line 38: I think that is unnecessary. If the bot would be in use in many places for many different repositories, it might make sense.

scripts/bot-build.slurm
```diff
@@ -32,12 +32,20 @@ else
...
+if [ ! -f ${CHECK_BUILD_SCRIPT} ] && [ -f ${CHECK_RESULT_SCRIPT} ]; then
+    echo "${CHECK_BUILD_SCRIPT} not found in ${PWD}, but ${CHECK_RESULT_SCRIPT} is found, so using that as fallback..."
+    CHECK_BUILD_SCRIPT=${CHECK_RESULT_SCRIPT}
```
trz42, line 41: Just rename from "RESULT" to "BUILD".

trz42 (dismissed):
OK to rename. No need to keep the old as a fall back.

scripts/bot-build.slurm
boegel: Well, we do need the corresponding change in `compatibility-layer` and `software-layer` (and in all branches...) first then?

I wouldn't mind having this safeguard in place for a couple of days to ease the transition, I don't see any downsides here.
I can open another PR right after this one to strip out this fallback mechanism again, so we don't forget about doing so...

bedroge:
Since Thomas is fine with a rename without a fallback, and he's probably the only one using this outside of EESSI, I'd say we just do a rename only.

scripts/bot-build.slurm
boegel: removed fallback in [4b013d3](https://github.com/EESSI/eessi-bot-software-layer/pull/189/commits/4b013d3c976425a81b7264bdcd0a09ec05cc9f37)

scripts/bot-build.slurm
boegel: fixed in [4b013d3](https://github.com/EESSI/eessi-bot-software-layer/pull/189/commits/4b013d3c976425a81b7264bdcd0a09ec05cc9f37)

bedroge (approved):

PR #198, opened 2023-08-07
[v0.1] code polishing: tools/logging.py

trz42:
Code polishing pass on `tools/logging.py` for release v0.1:

- imports: same structure (some sections/categories are empty), removed one unused import (`json` was needed for a removed function)
- docstrings: added args and returns
- comments: added a comment regarding fixed location+name of log file
- functions: removed unused function `log_event`

### DIFF
--- a/tools/logging.py
+++ b/tools/logging.py
@@ -25,6 +25,14 @@ LOG = os.path.join(os.getenv('HOME'), 'eessi-bot-software-layer.log')
 def error(msg, rc=1):
     """
     Print an error and exit
+
+    Args:
+        msg (string): error message to be printed
+        rc (int): error code
+
+    Returns:
+        does not return anything (function never returns, but rather exits the
+        program)
     """
     sys.stderr.write(msg + "\n")
     sys.exit(rc)
@@ -33,6 +41,12 @@ def error(msg, rc=1):
 def log(msg):
     """
     Log message
+
+    Args:
+        msg (string): error message to be printed
+
+    Returns:
+        does not return anything
     """
     with open(LOG, 'a') as fh:
         timestamp = datetime.datetime.now().strftime("%Y%m%d-T%H:%M:%S")
--- a/tools/logging.py
+++ b/tools/logging.py
@@ -11,7 +11,6 @@
 
 # Standard library imports
 import datetime
-import json
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


tools/logging.py
```diff
@@ -8,17 +8,31 @@
...
+
+# Standard library imports
 import datetime
-import json
 import os
 import sys
 
+# Third party imports (anything installed into the local Python environment)
+
+# Local application imports (anything from EESSI/eessi-bot-software-layer)
```
boegel, line 19: Should we include these lines if there are no such imports?
Maybe add a `# (none yet)` below each?

tools/logging.py
```diff
@@ -8,17 +8,31 @@
...
+
+# TODO possibly replace 'HOME' (and log file name) with a configurable value
```
boegel, line 21: Or just one variable to determine full path to log file, with `~/eessi-bot-software-layer.log` being the default is none is specified?

tools/logging.py
```diff
@@ -27,22 +41,13 @@ def error(msg, rc=1):
 def log(msg):
     """
     Log message
+
+    Args:
+        msg (string): error message to be printed
+
+    Returns:
+        does not return anything
```
boegel, line 49: Well, in fact, `None` is always implicitly returned... Not sure how the rephrase this

tools/logging.py
```diff
@@ -27,22 +41,13 @@ def error(msg, rc=1):
...
     """
     with open(LOG, 'a') as fh:
         timestamp = datetime.datetime.now().strftime("%Y%m%d-T%H:%M:%S")
         fh.write('[' + timestamp + '] ' + msg + '\n')
-
-
-def log_event(request):
```
boegel, line 36: note: totally unused, so OK to delete :+1:

tools/logging.py
trz42: I thought it would be good to keep them, so in the future if we add imports we put them in one of the categories. Will add the `# (none yet)`.

tools/logging.py
trz42: Turns out we already have a config setting for this ... or more precisely two:

```ini
[event_handler]
log_path = ...

[job_manager]
log_path = ...
```

We also have an issue to improve logging (#91). Will change the comment accordingly.

tools/logging.py
trz42: Replaced it with

```
        None (implicitly)
```

boegel (approved):

PR #199, opened 2023-08-07
[v0.1] code polishing: tools/config.py

trz42:
Code polishing pass on `tools/config.py` for release v0.1:

- imports: same structure (some sections/categories are empty)
- docstrings: slightly rephrased some + added a bit more information
- comments: added some clarifying comments + added a TODO for an unused argument

### DIFF
--- a/tools/config.py
+++ b/tools/config.py
@@ -52,10 +52,14 @@ def check_required_cfg_settings(req_settings, path="app.cfg"):
     Returns:
         does not return anything
     """
+    # TODO argument path is not being used
     cfg = read_config()
+    # iterate over keys in req_settings which correspond to sections ([name])
+    # in the configuration file (.ini format)
     for section in req_settings.keys():
         if section not in cfg:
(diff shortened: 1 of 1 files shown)


tools/config.py
```diff
@@ -8,18 +8,27 @@
 #
 # license: GPLv2
 #
+
+# Standard library imports
 import configparser
 import sys
 
+# Third party imports (anything installed into the local Python environment)
```
boegel, line 16: add `# (none yet)` to avoid confusion?

tools/config.py
```diff
@@ -34,17 +43,24 @@ def read_config(path='app.cfg'):
...
+    Reads the config file, checks if it contains the required settings,
+    if not logs an error message and exits.
+
     Args:
         req_settings (dict (str, list)): required settings
         path (string): path to the configuration file
+
     Returns:
-        None
+        does not return anything
```
boegel, line 54: not entirely correct, since there's an implicit return of `None`

boegel (approved):

PR #202, opened 2023-08-07
[v0.1] code polishing: tools/args.py

trz42:
Code polishing pass on `tools/args.py` for release v0.1:

- imports: added comments for categories of imports
- docstrings: added missing information

### DIFF
--- a/tools/args.py
+++ b/tools/args.py
@@ -9,7 +9,13 @@
 #
 # license: GPLv2
 #
+
+# Standard library imports
 import argparse
+
+# Third party imports (anything installed into the local Python environment)
+
+# Local application imports (anything from EESSI/eessi-bot-software-layer)
 
 
 def parse_common_args(args=None):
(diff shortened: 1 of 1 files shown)


tools/args.py
```diff
@@ -9,10 +9,26 @@
 #
 # license: GPLv2
 #
+
+# Standard library imports
 import argparse
 
+# Third party imports (anything installed into the local Python environment)
+
+# Local application imports (anything from EESSI/eessi-bot-software-layer)
```
boegel, line 18: add `# (none yet)` here and above to avoid confusion?

tools/args.py
```diff
@@ -9,10 +9,26 @@
...
+
 
 def parse_common_args(args=None):
+    """
+    Create an ArgumentParser instance, add common arguments and parse them
```
boegel, line 23: Reword as below:

```
Parse common arguments that are shared by event handler and job manager

boegel (changes requested):

boegel (approved):

Issue #213, opened 2023-09-08
fluke download failures


laraPPr:
Sometimes a build failes because EasyBuild crashes after it fails to download a file.

This happened a lot in the TensorFlow PR https://github.com/EESSI/software-layer/pull/321: 
- `7194`: who failed to get libpng-1.6.37.tar.gz
- `7200`, `7199`,...: There was also a problem with GitHub a few days which caused a lot of build failures 

it might be an option to implement a pre-build `fetch` phase. So that all the builds have the necessary sources.

```
Yes, that should probably become a feature in the bot, we can look into adding support for a pre-
build fetch phase for example, so we can instruct the bot to first try and fetch all sources before 
letting it submit build jobs.

Ideally the bot would then automatically first submit a single fetch job when it gets the instruction 
to build, but that may be a bit harder (since the fetch part could take a while, and we don't want to 
block the bot to until fetch is done).
```

boegel:
@trz42 Thoughts on this?
This could help significantly to avoid fluke build failures I think...

The bot could "prefetch" sources to a directory before launching any build jobs, so they don't need to download anything anymore; we also do this for `eb --job`.

trz42:
I understand the issue, but I am a unsure what a good solution to this would be.

1. If EasyBuild fails to download something, maybe it should have a means to retry? Like we added retry capability to some bot functions with an increasing delay between attempts.
2. I don't see how this could be easily implemented at the bot-side. The bot does not know much about what it actually processes (PR for some repository that provides a `bot/build.sh`).
3. If it would do some kind of pre-processing, the whole job management is getting significantly more complex than it is right now.

If the main issue is, that, sometimes, a job fails late just because some source cannot be downloaded, a relatively easy way around that might be to improve the handling of sources in `bot/build.sh` or even `EESSI-pilot-install-software.sh` by adding a `fetch` phase before the actual build phase. That fetching could be repeated a couple of times. Even if it never succeeds, it might consume fewer resources because that failure is raised before the actual building begins.

Note, I think because of 2 & 3 above, this is not easy to implement. Adjusted the difficulty label accordingly.

boegel:
The `shared_fs_path` configuration setting implemented in https://github.com/EESSI/eessi-bot-software-layer/pull/214 allows to largely fix this, by using a shared directory for `$EASYBUILD_SOURCEPATH`, see also https://github.com/EESSI/software-layer/pull/337 .

PR #214, opened 2023-09-21
add `shared_fs_path` configuration setting

boegel:
~WIP because I'm still testing whether this works as expected, see also companion PR https://github.com/EESSI/software-layer/pull/337~

Works as expected, testing along with https://github.com/EESSI/software-layer/pull/337 via https://github.com/boegel/software-layer/pull/27

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,22 @@ build_job_script = PATH_TO_EESSI_BOT/scripts/bot-build.slurm
 This points to the job script which will be submitted by the event handler.
 
 ```
+shared_fs_path = PATH_TO_SHARED_DIRECTORY
+```
+
+Via `shared_fs_path` the path to a directory on a shared filesystem (NFS, etc.) can be provided,
+which can be leveraged by the `bot/build.sh` script to store files that should be available across build jobs
+(software source tarballs, for example).
+
+```
+build_logs_dir = PATH_TO_BUILD_LOGS_DIR
+```
+
+If build logs should be copied to a particular (shared) directory under certain conditions,
+for example when a build failed, the `build_logs_dir` can be set to the path to which logs
+should be copied by the `bot/build.sh` script.
+
+```
 container_cachedir = PATH_TO_SHARED_DIRECTORY
 ```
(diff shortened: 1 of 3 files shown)


boegel:
@trz42 I feel the changes here are trivial enough to include them in the `v0.1.0` release of the bot, but I'll let you be the judge of that...

trz42:
> @trz42 I feel the changes here are trivial enough to include them in the `v0.1.0` release of the bot, but I'll let you be the judge of that...

Not fully clear to me what the tests in https://github.com/EESSI/software-layer/pull/337 and https://github.com/boegel/software-layer/pull/27 show ... there is very little information about what has been tested (e.g., a successful download and then the reuse of the downloaded sources; how does this improve the _fluke_ download issue if the first download for the $N-1$-th dependency fails?). Are the tests sufficient to demonstrate that this PR solves the issue (https://github.com/EESSI/eessi-bot-software-layer/issues/213) and what undesired consequences may they have (multiple jobs writing concurrently to the shared directory or worse the same files ...).

To me this looks more like a performance optimisation (avoiding repeated downloads of the same sources), than solving the _fluke_ download issue.

boegel:
You're right about the concurrent downloads to a shared place. The idea is to combine setting `shared_fs_path` with starting only a single build job per PR initially, which takes care of the downloading.
EasyBuild currently doesn't use a lock when a download is being done, but it could, and if a 2nd download waits for a while until the download lock is released, that would prevent problems with concurrent downloads.

It doesn't fully avoid the download fluke problem, so I wouldn't close https://github.com/EESSI/eessi-bot-software-layer/issues/213 just yet, but it should make it as lot less likely since every file only needs to be downloaded once.
I'm not sure we can fully avoid download problems btw, we're always at the mercy of whatever site provides the source tarballs?

Keep in mind that the same source tarball can be used for other PRs as well, like `zlib` 1.2.13 for example.
Right now source tarballs are *always* downloaded in a build job, there's no caching of downloads at all.

trz42 (approved):
Looks good. Should be fine.

PR #217, opened 2023-09-30
script to clean up tarballs of jobs given a PR number

trz42:
To delete all major tarballs (temporary storage and build artefacts such as software installations) of PR `NUMBER`, simply run with

```bash
./cleanup_pr.sh NUMBER
```

from bot directory that contains `app.cfg` (used to determine base directory for jobs, e.g., setting `jobs_base_dir`).

* Add `-D` or `--dry-run` to check what the script would do.
* Add `-b DIR` or `--jobs-base-dir DIR` where `DIR` points to the base directory of jobs.


cleanup_pr.sh
```diff
@@ -0,0 +1,156 @@
...
+      POSITIONAL_ARGS+=("$1") # save positional arg
+      shift
+      ;;
+  esac
+done
+
+# restore potentially parsed filename(s) into $*
+set -- "${POSITIONAL_ARGS[@]}"
+
+if [[ $# -lt 1 ]]; then
```
boegel, line 79: Don't we want exactly 1 PR number here?

If so, use `$# -ne 1`, and adjust the error message, like `"ERROR: Exactly one PR number should be provided"`.

cleanup_pr.sh
```diff
@@ -0,0 +1,156 @@
...
+echo
+echo "$([[ ${dry_run} = true ]] && echo "DRY_RUN: ")removing tmp storage tarballs for PR ${pull_request}"
+for d in $(find ${jobs_base_dir}/* -maxdepth 1 -type d -wholename */pr_${pull_request})
+do
+  for f in $(find $d -type f -wholename "*[0-9].tgz")
+  do
+    if ${dry_run} = true ; then
+      echo "DRY_RUN: rm '$f' ($(ls -lh $f | awk '{print $5}'))"
+    else
+      rm $f
```
boegel, line 130: maybe also emit a message when removing a file?

```
echo "Removing $f..."`

boegel (changes requested):

boegel:
@trz42 Next step (maybe in follow-up PR) could be to run this script every time a PR is closed/merged?

cleanup_pr.sh
trz42: Done in https://github.com/EESSI/eessi-bot-software-layer/pull/217/commits/1f9c0dab85e3c4cceb984a968753533ceeed7f84

cleanup_pr.sh
trz42: Done in https://github.com/EESSI/eessi-bot-software-layer/pull/217/commits/1f9c0dab85e3c4cceb984a968753533ceeed7f84

trz42:
> @trz42 Next step (maybe in follow-up PR) could be to run this script every time a PR is closed/merged?

Good idea.

We could implement this as follows:
- When PR is closed/merged, submit a new job that executes `cleanup_pr.sh` (or `scripts/bot-cleanup_pr.slurm`) with some additional parameters.
- Additional parameters could be
  - `-p|--cleanup-policy POLICY` with policies `all` -- cleans up all directories for the PR (doesn't run target repository-specific cleanup script), `repo_defined` -- for each job runs `bot/cleanup_pr.sh` provided by the target repository, `tarballs` -- for each job removes all tarballs (suffix `.tar.gz` or `.tgz`), (possibly) distinguish different types of tarballs for what was built or what contains temporary storage, `none` -- doesn't clean up anything; policy should be defined in `app.cfg`, default policy could be `none` or `all`
  - `-j|--jobid JOBID` only cleans up directory for the given jobid
- The target repository could provide a `bot/cleanup_pr.sh` which cleans up a directory for a given directory (or current directory).
- (Optionally) Having these scripts would make the implementation of a command such as `bot: cleanup job:JOBID` or just `bot: cleanup` or `bot: cleanup storage:tmp` relatively easy.

boegel:
@trz42 Should we move the script into the `scripts` subdirectory?

boegel (approved):

PR #218, opened 2023-10-04
release notes for v0.2.0

trz42:
Just describes the addition and purpose of the new script `scripts/cleanup_pr.sh`.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,6 +1,14 @@
 This file contains a description of the major changes to the EESSI
 build-and-deploy bot. For more detailed information, please see the git log.
 
+v0.2.0 (4 october 2023)
+--------------------------
+
+* new script `scripts/cleanup_pr.sh` that removes tarballs (temporary storage and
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES
```diff
@@ -1,6 +1,14 @@
 This file contains a description of the major changes to the EESSI
 build-and-deploy bot. For more detailed information, please see the git log.
 
+v0.2.0 (4 october 2023)
```
boegel, line 4: Since this is a minor enhancement (bot itself hasn't changed, it's just an extra script which is currently totally disconnected from the bot), I'd use `0.1.1`, but I guess that's debatable (since it's not a bug fix either)

We need a 3rd vote... @bedroge ?

trz42: We could argue that not cleaning up was a bug???

trz42:
Agreed that it is too small change to warrant a new release. We can also just copy the cleanup script to where we need it until it is available via a release.

PR #220, opened 2023-10-04
omit header in squeue output

### LINKED ISSUE #219: Issue parsing the output of squeue
The current setup of the `job_manager` assumes that when the command `squeue` is run there will be 1 line for the header and than the list of running jobs. This can however differ. On the systems at the UGent the header takes up 2 lines. 

When starting the job_manager on the UGent system we go the following error:

```
run_cmd(): Error running '/usr/bin/scontrol --oneliner show jobid JOBID' in 'None
           stdout ''
           stderr 'scontrol_print_job error: Invalid job id specified
'
           exit code 1
```

This is the output of `squeue` on the Ugent system:

```
CLUSTER: donphan
             JOBID PARTITION     NAME     USER    STATE       TIME TIME_LIMI  NODES NODELIST(REASON)
          20010031   donphan INTERACT vsc46128  RUNNING    3:56:49   5:00:00      1 node4006.donphan.os
```

The current setup of the `job_manager` will skip and not parse the first line but not the second. So the `job_manager` crashed when trying to look at the job with id `JOBID`.

trz42:
Adds `--noheader` argument to `squeue` command and then parses all output lines to determine current jobs.

Fixes #219

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -106,7 +106,7 @@ class EESSIBotSoftwareLayerJobManager:
         if username is None:
             raise Exception("Unable to find username")
 
-        squeue_cmd = "%s --long --user=%s" % (self.poll_command, username)
+        squeue_cmd = "%s --long --noheader --user=%s" % (self.poll_command, username)
         squeue_output, squeue_err, squeue_exitcode = run_cmd(
             squeue_cmd,
             "get_current_jobs(): squeue command",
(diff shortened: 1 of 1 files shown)


eessi_bot_job_manager.py
```diff
@@ -125,9 +125,9 @@ def get_current_jobs(self):
         }
 
         # get job info, logging any Slurm issues
-        # Note, the first two lines of the output are skipped ("range(2,...)")
-        #     because they contain header information.
-        for i in range(2, len(lines)):
+        # Note, all output lines of squeue are processed because we run it with
+        # --noheader.
+        for i in range(0, len(lines)):
```
boegel, line 130: Wouldn't `for line in lines` be cleaner then?

boegel (changes requested):

eessi_bot_job_manager.py
trz42: Yep.

trz42: Done in https://github.com/EESSI/eessi-bot-software-layer/pull/220/commits/448dcfadc14cbaf14b923bff22b5a238dcfce019

boegel (approved):

Issue #223, opened 2023-10-19
Newest version of PyGithub is not compatible with the bot


laraPPr:
PyGithub recently realease PyGithub 2.1.1 which is no longer compatible with the bot. The last version that of PyGithub that is compatible with the bot is v1.59.1. [link](https://pypi.org/project/PyGithub/#history)

With PyGithub 2.1.1 the bot has the following error when trying to submit more than one comment to a pr (It can respond to one bot command but when a seccond command is submitted in the pr the bot cannot respond to it).

The error: 

```
Traceback (most recent call last):
  File "/kyukon/scratch/gent/461/vsc46128/EESSI/venv_eessi_bot_login_p38/lib/python3.8/site-packages/pyghee/lib.py", line 170, in process_event
    self.handle_event(event_info, log_file=log_file)
  File "/kyukon/scratch/gent/461/vsc46128/EESSI/venv_eessi_bot_login_p38/lib/python3.8/site-packages/pyghee/lib.py", line 102, in handle_event
    handler(event_info, log_file=log_file)
  File "/kyukon/scratch/gent/461/vsc46128/EESSI/eessi-bot-software-layer/eessi_bot_event_handler.py", line 208, in handle_issue_comment_event
    issue_comment = create_comment(repo_name, pr_number, comment_body)
  File "/kyukon/scratch/gent/461/vsc46128/EESSI/eessi-bot-software-layer/tools/pr_comments.py", line 45, in create_comment
    gh = github.get_instance()
  File "/kyukon/scratch/gent/461/vsc46128/EESSI/eessi-bot-software-layer/connections/github.py", line 104, in get_instance
    if not _gh or (_token and datetime.datetime.utcnow() > _token.expires_at):
TypeError: can't compare offset-naive and offset-aware datetimes
```

We should try to resolve this error so that the bot uses the latest version of PyGitlab. 

If we implement the fix we should also update `requirements.txt` so that we at least require PyGithub version 2.1.1.

laraPPr:
Fixed by changing the datetime. But requires PyGithub to be at least 2.1.1

```
from datetime import datetime, timezone
if not _gh or (_token and datetime.now(timezone.utc) > _token.expires_at):
```

boegel:
Seems like the problems we're seeing are caused by a documented breaking change in PyGithub v2.1.0, see https://github.com/PyGithub/PyGithub/releases/tag/v2.1.0.post0

[...]

PR #224, opened 2023-10-20
Upgrade bot to be compatible with PyGithub v2.1.1

laraPPr:
I tested the changes with python 3.8

resolves the following issue: https://github.com/EESSI/eessi-bot-software-layer/issues/223

### DIFF
--- a/connections/github.py
+++ b/connections/github.py
@@ -11,7 +11,7 @@
 #
 
 # Standard library imports
-import datetime
+from datetime import datetime, timezone
 import time
 
 # Third party imports (anything installed into the local Python environment)
@@ -101,7 +101,7 @@ def get_instance():
     global _gh, _token
     # TODO Possibly renew token already if expiry date is soon, not only
     #      after it has expired.
-    if not _gh or (_token and datetime.datetime.utcnow() > _token.expires_at):
+    if not _gh or (_token and datetime.now(timezone.utc) > _token.expires_at):
         _gh = connect()
     return _gh
 
--- a/connections/github.py
+++ b/connections/github.py
@@ -103,10 +103,10 @@ def get_instance():
     #      after it has expired.
 
     # Check if PyGithub version is < 1.56
-    if hasattr(github, 'GithubRetry' is False:
+    if hasattr(github, 'GithubRetry') is False:
         if not _gh or (_token and datetime.utcnow() > _token.expires_at):
             _gh = connect()
-    elif hasattr(github, 'GithubRetry' is True:
+    elif hasattr(github, 'GithubRetry') is True:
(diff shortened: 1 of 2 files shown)


boegel:
PyGithub 2.1.1 requires Python >= 3.7, so applying this fix implies that the bot will no longer support Python 3.6 (see also failing CI)...

That's painful because Python 3.6 is the default Python in RHEL 8 & derivates, so ideally we retain compatibility with that.

I wonder if there's a smart way in which can check whether or not we need to compare with a `datetime` instance that includes `tzinfo`, or not.
Maybe we can easily check which version of PyGithub we're dealing with?

boegel:
Hmm, I can't find a direct way to check the PyGithub version, but we can do it by proxy: the `` class was introduced in PyGithub 2.x, so:

```
[bot@login1 ~]$ python3.6 -m pip list | grep PyGithub
PyGithub (1.56)
[bot@login1 ~]$ python3.6 -c "import github; print(hasattr(github, 'GithubRetry'))"
False
```

```
[boegel@login1 ~]$ python3.9 -m pip list | grep PyGithub
PyGithub           2.1.1
[boegel@login1 ~]$ python3.9 -c "import github; print(hasattr(github, 'GithubRetry'))"
True

File: connections/github.py [hunk 1]
```diff
@@ -101,8 +101,14 @@ def get_instance():
...
-    if not _gh or (_token and datetime.datetime.utcnow() > _token.expires_at):
-        _gh = connect()
+
+    # Check if PyGithub version is < 1.56
+    if hasattr(github, 'GithubRetry') is False:
+        if not _gh or (_token and datetime.utcnow() > _token.expires_at):
+            _gh = connect()
+    elif hasattr(github, 'GithubRetry') is True:
+        if not _gh or (_token and datetime.now(timezone.utc) > _token.expires_at):
+            _gh = connect()
```
boegel, line 111: `if ... is False` is kind of silly, and positive logic is easier to follow, and there's a lot of copy-pasting, so I would change this to:

```python
    # PyGithub 2.x expects a datetime object with timezone info,
    # PyGithub 1.x requires one without timezone info;
    # see also https://github.com/PyGithub/PyGithub/pull/2565
    if hasattr(github, 'GithubRetry'):
        # PyGithbub 2.x
        time_now = datetime.now(timezone.utc)
    else:
        # PyGithbub 1.x
        time_now = datetime.utcnow()

    if not _gh or (_token and time_now > _token.expires_at):
        _gh = connect()

boegel (changes requested):

boegel (approved):
lgtm

trz42:
Did some test runs with Python/3.6.8 and Python/3.9.16. See https://github.com/trz42/software-layer/pull/67

Looks good.

PR #231, opened 2023-11-25
recommend smee container available at `ghcr.io/probot/smee.io` in `README` (which works on `x86_64` + `aarch64`)

boegel:
see also https://github.com/probot/smee.io

fix for:

```
FATAL:   While checking container encryption: could not open image /home/boegel/eessi-bot-software-layer/smee-client_latest.sif:
the image's architecture (amd64) could not run on the host's (arm64)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,15 +35,15 @@ On the `bot machine` we need a tool which receives events relayed from
 client for this. The Smee client can be run via a container as follows
 
 ```
-singularity pull docker://deltaprojects/smee-client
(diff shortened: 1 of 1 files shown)


boegel:
Ugh, nevermind, same problem with the container from `ghcr.io/probot/smee.io` 🤦

boegel:
I worked around this by installing `smee` in the current directory with:

```
npm install smee-client
```

and then running it with:
```
node_modules/smee-client/bin/smee.js --url ...
```

PR #237, opened 2024-01-25
New bot command `bot: status`

laraPPr:
The command creates a comment that contains a table with all the builds done in the pr where the command is called. 

example: 
- https://github.com/laraPPr/software-layer/pull/4#issuecomment-1910007732

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -25,7 +25,7 @@ import waitress
 # Local application imports (anything from EESSI/eessi-bot-software-layer)
 from connections import github
 import tasks.build as build
-from tasks.build import check_build_permission, get_architecture_targets, get_repo_cfg, submit_build_jobs, request_bot_build_issue_comments
+from tasks.build import check_build_permission, get_architecture_targets, get_repo_cfg, \    submit_build_jobs, request_bot_build_issue_comments
 import tasks.deploy as deploy
 from tasks.deploy import deploy_built_artefacts
 from tools import config
@@ -493,12 +493,16 @@ class EESSIBotSoftwareLayer(PyGHee):
         repo_name = event_info['raw_request_body']['repository']['full_name']
         pr_number = event_info['raw_request_body']['issue']['number']
         status_table = request_bot_build_issue_comments(repo_name, pr_number)
-        
+
         comment = f"This is the status of all the `bot: build` commands:"
         comment += f"\n|arch|result|date|status|url|"
         comment += f"\n|----|------|----|------|---|"
-        for x in range(0,len(status_table['date'])):
-            comment += f"\n|{status_table['arch'][x]}|{status_table['result'][x]}|{status_table['date'][x]}|{status_table['status'][x]}|{status_table['url'][x]}|"
+        for x in range(0, len(status_table['date'])):
+            comment += f"\n|{status_table['arch'][x]}|"
+            comment += f"{status_table['result'][x]}|"
+            comment += f"{status_table['date'][x]}|"
+            comment += f"{status_table['status'][x]}|"
+            comment += f"{status_table['url'][x]}|"
 
         self.log(f"PR opened: comment '{comment}'")
         repo = gh.get_repo(repo_name)
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -509,7 +509,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         self.log(f"PR opened: comment '{comment}'")
         repo = gh.get_repo(repo_name)
         pull_request = repo.get_pull(pr_number)
-        issue_comment = pull_request.create_issue_comment(comment)
+        issue_comment = pull_request.create_issue_comment(comment_status)
         return issue_comment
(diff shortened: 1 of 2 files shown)


tasks/build.py
```diff
@@ -760,3 +760,65 @@ def check_build_permission(pr, event_info):
...
+            # iterate through the comments to find the one where the status of the build was in
+            if config.read_config()["submitted_job_comments"]['initial_comment'][:20] in comment['body']:
+
+                # get archictecture from comment['body']
+                first_line = comment['body'][:comment['body'].find('/n')]
+                arch_map = get_architecture_targets(cfg)
+                for arch in arch_map.keys():
+                    target_arch = '/'.join(arch.split('/')[1:])
+                    if target_arch in first_line:
+                        status_table['arch'].append(target_arch)
```
trz42, line 788: So this iterates over all defined architectures and adds it to `status_table['arch']` if it matches what it found in the first line of the comment. Alternatively, one could also take the value directly from comment?

tasks/build.py
```diff
@@ -760,3 +760,65 @@ def check_build_permission(pr, event_info):
...
+
+                # add date, status, url to  status_table if
+                for value in values:
+                    if value['job status'] == 'finished':
+                        status_table['date'].append(value['date'])
+                        status_table['status'].append(value['job status'])
+                        status_table['url'].append(comment['html_url'])
+                        if 'FAILURE' in value['comment']:
+                            status_table['result'].append(':cry: FAILURE')
+                        elif 'SUCCES' in value['comment']:
```
trz42: Minor typo?
```suggestion
                        elif 'SUCCESS' in value['comment']:
```

tasks/build.py
```diff
@@ -760,3 +760,65 @@ def check_build_permission(pr, event_info):
     else:
         log(f"{fn}(): GH account '{build_labeler}' is authorized to build")
         return True
+
+
+def request_bot_build_issue_comments(repo_name, pr_number):
+    status_table = {'arch': [], 'date': [], 'status': [], 'url': [], 'result': []}
```
trz42, line 766: Wonder a bit if it would be better to let `status_table` be a list of dictionaries instead of a dictionary of lists where we assume that lists are always identically formed (same length, elements at same index belong to each other).

tasks/build.py
```diff
@@ -760,3 +760,65 @@ def check_build_permission(pr, event_info):
...
+                for arch in arch_map.keys():
+                    target_arch = '/'.join(arch.split('/')[1:])
+                    if target_arch in first_line:
+                        status_table['arch'].append(target_arch)
+
+                # get date, status, url and result from the markdown table
+                comment_table = comment['body'][comment['body'].find('|'):comment['body'].rfind('|')+1]
+
+                # Convert markdown table to a dictionary
+                lines = comment_table.split('\n')
```
trz42, line 794: Maybe `split('\n')` could also be used to get `first_line` (see above).

trz42 (changes requested):
Looks great. Some small comments. The result is for sure very useful!!!

eessi_bot_event_handler.py
```diff
@@ -476,6 +477,40 @@ def handle_bot_command_show_config(self, event_info, bot_command):
         issue_comment = self.handle_pull_request_opened_event(event_info, pr)
         return f"\n  - added comment {issue_comment.html_url} to show configuration"
 
+    def handle_bot_command_status(self, event_info, bot_command):
+        """
+        Handles bot command 'status' by running the handler for events of
+        type pull_request with the action opened.
```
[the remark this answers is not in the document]
laraPPr: forgot to update this fully. copied it from one of the other funtions. will fix it.

tasks/build.py
laraPPr: yes but than you first make it neccessary for 
``` 
initial_comment = New job on instance `{app_name}` for architecture `{arch_name}` for repository `{repo_id}` in job dir `{symlink}` 
``` 
to stay exactly the same now even if minor changes are made this will keep working.
Now I have tried to right this function in such a way that it allows changes to `initial_comment` in `app.cfg`

tasks/build.py
laraPPr: I'm open to that. Because it actually takes some effort to put it in a dictionary with lists.

eessi_bot_event_handler.py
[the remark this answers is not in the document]
laraPPr: updated

tasks/build.py
```diff
@@ -760,3 +760,65 @@ def check_build_permission(pr, event_info):
     else:
         log(f"{fn}(): GH account '{build_labeler}' is authorized to build")
         return True
+
+
+def request_bot_build_issue_comments(repo_name, pr_number):
```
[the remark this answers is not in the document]
laraPPr: updated

tasks/build.py
trz42: Ack. We can think about that later in case it becomes necessary.

trz42 (approved):
Great improvement @laraPPr !

PR #247, opened 2024-02-09
add function that reads PR section from job metadata and use it in job manager

trz42:
This is a small change moving code that is used in `eessi_bot_job_manager.py` into a function defined in `tools/job_metadata.py` and then use that function in `eessi_bot_job_manager.py`.

It's a step towards syncing the codes of the NESSI and EESSI bots.

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -42,7 +42,7 @@ from pyghee.utils import log
 from connections import github
 from tools import config, run_cmd
 from tools.args import job_manager_parse
-from tools.job_metadata import read_metadata_file
+from tools.job_metadata import read_job_metadata_from_file, read_metadata_file
 from tools.pr_comments import get_submitted_job_comment, update_comment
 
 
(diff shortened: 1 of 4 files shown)


eessi_bot_job_manager.py
```diff
@@ -253,7 +253,7 @@ def determine_finished_jobs(self, known_jobs, current_jobs):
 
     def read_job_pr_metadata(self, job_metadata_path):
         """
-        Read job metadata file and return the contents of the 'PR' section.
+        Determine metadata od a job
```
bedroge, line 256: This function now only calls another function, so shouldn't we just remove it and replace all calls to it by a call to `read_job_metadata_from_file`?

trz42: Good idea. Implemented in https://github.com/EESSI/eessi-bot-software-layer/pull/247/commits/7ed3d8d32b24da271bbd0ac0e7b513f1e6ee2b9e

Moved corresponding test from `tests/test_eessi_bot_job_manager.py` to `tests/test_tools_job_metadata.py`.

bedroge (approved):
Lgtm

PR #248, opened 2024-02-11
Add a comment when downloading and merging from a PR fails via `comment_download_pr()`

Neves-P:
Adds a new function `comment_download_pr()` that takes the exit codes from `download_pr()` to give an helpful message via a GH comment on an open PR. 

The aim is to mostly handle situations where a merge conflict due to an upstream merge prevents the `git apply` step in `download_pr()` from working. At this stage of `download_pr()`, the recent changes (a patch) are downloaded and `git apply` is used. When a merge conflict is present, `git apply` fails with an error and which not reported by the bot and made visible to the user.

For this change, I also explicitly pass the error codes out of `download_pr()` to be able to expose the error message to the user and handle the errors.

To be done in the future, add unit tests to the new functionality. I have started, but not managed to finish the mocking and testing framework for this function. Adding this PR so we have have this functionality earlier.

### DIFF
--- a/tasks/build.py
+++ b/tasks/build.py
@@ -464,9 +464,9 @@ def prepare_jobs(pr, cfg, event_info, action_filter):
                                                                                        pr, job_dir)
 
             if download_pr_exit_code != 0:
-                download_comment = (f"{download_pr_error}"
-                                    f"\nUnable to download or merge changes between source branch and the destination branch.\n"
-                                    f"\nTip: This can usually be resolved by syncing your branch and solving any merge conflics")
+                download_comment = (f"`{download_pr_error}`"
+                                    f"\nUnable to download or merge changes between the source branch and the destination branch.\n"
+                                    f"\nTip: This can usually be resolved by syncing your branch and solving any merge conflics.")
                 download_comment = pr_comments.create_comment(repo_name=base_repo_name, pr_number=pr.number, comment=download_comment)
                 if download_comment:
                     log(f"{fn}(): created PR issue comment with id {download_comment.id}")
--- a/tasks/build.py
+++ b/tasks/build.py
@@ -345,21 +345,28 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
     # - 'curl' diff for pull request
     # - 'git apply' diff file
     git_clone_cmd = ' '.join(['git clone', f'https://github.com/{repo_name}', arch_job_dir])
-    clone_output, clone_error, clone_exit_code = run_cmd(git_clone_cmd, "Clone repo", arch_job_dir)
+    clone_output, clone_error, clone_exit_code = run_cmd(git_clone_cmd, "Clone repo", arch_job_dir, raise_on_error=False)
+    if clone_exit_code != 0:
+        return clone_output, clone_error, clone_exit_code
 
     git_checkout_cmd = ' '.join([
         'git checkout',
         branch_name,
     ])
     checkout_output, checkout_err, checkout_exit_code = run_cmd(git_checkout_cmd,
-                                                                "checkout branch '%s'" % branch_name, arch_job_dir)
+                                                                "checkout branch '%s'" % branch_name, arch_job_dir, raise_on_error=False)
+    if checkout_exit_code != 0:
+        return  checkout_output, checkout_err, checkout_exit_code
 
     curl_cmd = f'curl -L https://github.com/{repo_name}/pull/{pr.number}.diff > {pr.number}.diff'
-    curl_output, curl_error, curl_exit_code = run_cmd(curl_cmd, "Obtain patch", arch_job_dir)
-
+    curl_output, curl_error, curl_exit_code = run_cmd(curl_cmd, "Obtain patch", arch_job_dir, raise_on_error=False)
+    if curl_exit_code != 0:
+        return  curl_output, curl_error, curl_exit_code
+    
     git_apply_cmd = f'git apply {pr.number}.diff'
-    git_apply_output, git_apply_error, git_apply_exit_code = run_cmd(git_apply_cmd, "Apply patch", arch_job_dir)
-
+    git_apply_output, git_apply_error, git_apply_exit_code = run_cmd(git_apply_cmd, "Apply patch", arch_job_dir, raise_on_error=False)
+    if git_apply_exit_code != 0:
+        return git_apply_output, git_apply_error, git_apply_exit_code
 
 def apply_cvmfs_customizations(cvmfs_customizations, arch_job_dir):
     """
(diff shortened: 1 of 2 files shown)


Neves-P:
I'll handle all the linter comments tomorrow (12/02)

tasks/build.py
```diff
@@ -346,21 +346,60 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
     # - 'curl' diff for pull request
     # - 'git apply' diff file
     git_clone_cmd = ' '.join(['git clone', f'https://github.com/{repo_name}', arch_job_dir])
-    clone_output, clone_error, clone_exit_code = run_cmd(git_clone_cmd, "Clone repo", arch_job_dir)
+    print(f'cloning with command {git_clone_cmd}')
```
trz42, line 349: This prints to `stdout`. Should it maybe go to a log file?

tasks/build.py
```diff
@@ -346,21 +346,60 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
...
+    clone_output, clone_error, clone_exit_code = run_cmd(git_clone_cmd, "Clone repo", arch_job_dir, raise_on_error=False)
+    if clone_exit_code != 0:
+        return clone_output, clone_error, clone_exit_code
 
     git_checkout_cmd = ' '.join([
         'git checkout',
         branch_name,
     ])
+    print(f'checking out with command {git_checkout_cmd}')
```
trz42, line 358: This prints to `stdout`. Should it maybe go to a log file?

tasks/build.py
```diff
@@ -346,21 +346,60 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
...
     checkout_output, checkout_err, checkout_exit_code = run_cmd(git_checkout_cmd,
-                                                                "checkout branch '%s'" % branch_name, arch_job_dir)
+                                                                "checkout branch '%s'" % branch_name, arch_job_dir, raise_on_error=False)
+    if checkout_exit_code != 0:
+        return  checkout_output, checkout_err, checkout_exit_code
 
     curl_cmd = f'curl -L https://github.com/{repo_name}/pull/{pr.number}.diff > {pr.number}.diff'
-    curl_output, curl_error, curl_exit_code = run_cmd(curl_cmd, "Obtain patch", arch_job_dir)
-
+    print(f'curl with command {curl_cmd}')
```
trz42, line 365: This prints to `stdout`. Should it maybe go to a log file?

tasks/build.py
```diff
@@ -346,21 +346,60 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
...
+    curl_output, curl_error, curl_exit_code = run_cmd(curl_cmd, "Obtain patch", arch_job_dir, raise_on_error=False)
+    if curl_exit_code != 0:
+        return  curl_output, curl_error, curl_exit_code
+    
     git_apply_cmd = f'git apply {pr.number}.diff'
-    git_apply_output, git_apply_error, git_apply_exit_code = run_cmd(git_apply_cmd, "Apply patch", arch_job_dir)
+    print(f'git apply with command {git_apply_cmd}')
```
trz42, line 371: This prints to `stdout`. Should it maybe go to a log file?

tasks/build.py
```diff
@@ -346,21 +346,60 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
...
+    Return:
+        None (implicitly). A comment is created in the appropriate PR.
+        
+    """
+    if download_pr_exit_code != 0:
+        fn = sys._getframe().f_code.co_name
+        download_comment = (f"`{download_pr_error}`"
+                           f"\nUnable to download or merge changes between the source branch and the destination branch.\n"
+                           f"\nTip: This can usually be resolved by syncing your branch and resolving any merge conflicts.")
+        download_comment = pr_comments.create_comment(repo_name=base_repo_name, pr_number=pr.number, comment=download_comment)
```
trz42, line 397: This creates a new comment (for each job that failed), right? So we might get many new comments. Wonder if it could be possible to pass the result back to the calling function such that it is included in the responses to the `bot: build` command(s)?

trz42 (changes requested):
Nice work and more than just checking the result of `git apply`.

For some `print` statements, `log` might be better.

The calling function may not easily know which step (cloning the repo, branching, downloading the diff or applying the diff) failed. If the result would contain a 4th item one might include that information and use it when checking the exit code. Thus the note to the user could be adjusted and also the "Tip". Alternatively, one might only check the result of `git apply`.

Would the comment be created for all potential jobs or just once (because an exception is raised)? If multiple comments would be created, might be ok for now as it is already an improvement.

tasks/build.py
Neves-P: Absolutely, I've changed this and other occurrences to use `log()`.

tasks/build.py
Neves-P: That's a good point. I actually meant for it to just comment once and stop with the exception, but I did not actually verify that it works that way because when I was playing around with it I was only building for one target: https://github.com/Neves-Bot/software-layer/pull/12#issuecomment-1814142716

In that case of course the bot could only write one comment for that one target, so I'm completely sure the raising the error is enough 🤔

Neves-P:
> Nice work and more than just checking the result of `git apply`.

Thank you! I've address the other change requests individually.

> The calling function may not easily know which step (cloning the repo, branching, downloading the diff or applying the diff) failed. If the result would contain a 4th item one might include that information and use it when checking the exit code. Thus the note to the user could be adjusted and also the "Tip". Alternatively, one might only check the result of `git apply`.

Indeed, that's a great point, I've tried addressing that just now, where the error reporting and tip are handled depending on where the error occurs in `download_pr()`. This required returning  a new `error_stage` variable to inform `comment_download_pr()` of what to do. There might be a more clever way to do this, but I couldn't think of anything simple at first thought.

One very important note: I am in the process of reinstalling the bot on our HPC and setting it up to build for all the CPU targets we need in Groningen, so until that is done I can't test this at the moment as I was doing here: https://github.com/Neves-Bot/software-layer/pull/12. I know the code before the latest commits was working, but I can't say the same for this revision. I hope to have some time do finish that this week and then I'll already test this code. Unless there is a way of testing this outside of production code that I'm not aware of, it may be wise to wait on the merge until then :)

tasks/build.py
```diff
@@ -346,20 +348,94 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
     # - 'curl' diff for pull request
     # - 'git apply' diff file
     git_clone_cmd = ' '.join(['git clone', f'https://github.com/{repo_name}', arch_job_dir])
-    clone_output, clone_error, clone_exit_code = run_cmd(git_clone_cmd, "Clone repo", arch_job_dir)
+    log(f'cloning with command {git_clone_cmd}')
+    clone_output, clone_error, clone_exit_code = run_cmd(
+        git_clone_cmd, "Clone repo", arch_job_dir, raise_on_error=False
+        )
+    if clone_exit_code != 0:
+        error_stage = 'git clone'
```
trz42, line 356: Very good. Minor suggestion
```suggestion
        error_stage = ERROR_GIT_CLONE
```

and at the beginning (lines 34-79) add something like
```python
ERROR_GIT_CLONE = 'git clone'
```

tasks/build.py
```diff
@@ -346,20 +348,94 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
...
+        error_stage (string): a string informing the stage where download_pr() failed. Can be 'git clone',
+            'git checkout', 'curl', or 'git apply'.
+
+    Return:
+        None (implicitly). A comment is created in the appropriate PR.
+
+    """
+    if download_pr_exit_code != 0:
+        fn = sys._getframe().f_code.co_name
+        if error_stage == 'git clone':
```
trz42, line 409: Use globally defined variable `ERROR_GIT_CLONE`, e.g.,
```suggestion
        if error_stage == ERROR_GIT_CLONE:
```

trz42 (changes requested):
Looks great. Just a small suggestion on how to use string constants, e.g., define a variable at the beginning, then use that variable.

The text snippets that are used to print error messages could also be made configurable. That would allow us to change them in the configuration file (instead of changing them in the bot code). What do you think, would this make sense or would it be overkill here?

tasks/build.py
Neves-P: Done, for every occurrence of `error_stage`.

tasks/build.py
Neves-P: Moved string constants to globally defined variables in header.

Neves-P:
Thanks for the review! I've tried addressing them all, the string constants are globally defined at the top now and I've moved the text snippets to the `cfg` file. Let me know if something isn't right or if I should make more changes 👍

trz42 (approved):
Really nice work @Neves-P !!! Apologies for so many requests for changes.

PR #249, opened 2024-02-12
add function to determine comment id for a job and use it

trz42:
This is a follow up PR to sync bot codes used for EESSI and NESSI. It includes the following changes

- adds function `determine_pr_comment_id` which reads the attribute `pr_comment_id` from the job's metadata file (all current bot instances already provide that attribute when creating the job's metadata file)
- the function is then used when successful jobs are determined (in function `determine_successful_jobs`) and the comment id is added to a dictionary that contains information for each job
- another function (`determine_tarballs_to_deploy`) adds the comment id to a dictionary storing information about jobs/tarballs to deploy
- also includes slight code readability improvement by replacing a one letter variable name (`s`) with a more descriptive name (`job`)

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -26,6 +26,7 @@ from connections import github
 from tasks.build import CFG_DIRNAME, JOB_CFG_FILENAME, JOB_REPO_ID, JOB_REPOSITORY
 from tasks.build import get_build_env_cfg
 from tools import config, pr_comments, run_cmd
+from tools.job_metadata import read_job_metadata_from_file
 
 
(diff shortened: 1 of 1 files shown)


bedroge (approved):

tasks/deploy.py
```diff
@@ -74,6 +75,26 @@ def determine_job_dirs(pr_number):
     return job_directories
 
 
+def determine_pr_comment_id(job_dir):
```
bedroge, line 78: It would probably be good to implement a test for this function, but I see that we don't have any tests for the functions in `tasks/deploy.py`, so this can be done later for all of them.

PR #251, opened 2024-02-13
fix for `bot: status` was calling wrong values which caused the `bot: status` to fail

laraPPr:


tasks/build.py
```diff
@@ -794,7 +794,7 @@ def request_bot_build_issue_comments(repo_name, pr_number):
                 first_line = comment['body'].split('\n')[0]
                 arch_map = get_architecture_targets(cfg)
                 for arch in arch_map.keys():
-                    target_arch = '/'.join(arch.split('/')[1:])
+                    target_arch = '/'.join(arch.split('/')[-1:])
```
trz42, line 797: This might be a tiny bit simpler, right?
```suggestion
                    target_arch = '/'.join(arch.split('/')[-1])
```

trz42 (changes requested):
Small change, big effect. Looks good. No problem if you don't like the suggestion for improvement.

trz42 (approved):
Looks good.

PR #255, opened 2024-02-21
fix for job manager crash: Unable to contact slurm controller

laraPPr:


laraPPr:
Just was thinking that this pr might cause the bot to think that a job has finished when the squeue command fails.

laraPPr:
So this fix will keep the bot from crashing but the output of the bot in the pr could be wrong if slurm crashes when their the bot is handdling jobs at that moment.

eessi_bot_job_manager.py
```diff
@@ -113,8 +113,15 @@ def get_current_jobs(self):
             squeue_cmd,
             "get_current_jobs(): squeue command",
             log_file=self.logfile,
+            raise_on_error=False,
         )
 
+        if squeue_exitcode != 0:
+            current_jobs = {}
```
laraPPr, line 120:
```suggestion
            current_jobs = {'slurm manager failed'} 
```

Maybe we can do this and than we can first check for `determine_running_jobs`, `determine_new_jobs`, `determine_finished_jobs` if the `current_jobs` holds running jobs or if has nothing because the slurm was not availabel.

trz42 (changes requested):
I'd put the line https://github.com/laraPPr/eessi-bot-software-layer/blob/d458da17fd20a1aaed17132e929c96cb847b2ca5/eessi_bot_job_manager.py#L684 into a `try-except` block. In the `except` clause I'd just let the bot end the iteration with code as in https://github.com/laraPPr/eessi-bot-software-layer/blob/d458da17fd20a1aaed17132e929c96cb847b2ca5/eessi_bot_job_manager.py#L740..L746 and then let it `continue` to the next iteration in the main loop (maybe with an entry in the log file).

eessi_bot_job_manager.py
```diff
@@ -667,14 +674,26 @@ def main():
     if max_iter != 0:
         known_jobs = job_manager.get_known_jobs()
     while max_iter < 0 or i < max_iter:
+        # sleep poll_interval seconds (not for the first iteration)
+        if i != 0 and (max_iter < 0 or i < max_iter):
```
trz42, line 671: Could we simplify the logical expression to the following?
```suggestion
        if i != 0:
```
Should be ok because the loop body is only entered if `max_iter < 0 or i < max_iter` evaluates to `True`.

eessi_bot_job_manager.py
```diff
@@ -667,14 +674,26 @@ def main():
...
         log(
             "job manager main loop: known_jobs='%s'" % ",".join(
                 known_jobs.keys()),
             job_manager.logfile,
         )
 
-        current_jobs = job_manager.get_current_jobs()
+        try:
+            current_jobs = job_manager.get_current_jobs()
+        except:
```
trz42: I think, `run_cmd` raises a `RuntimeError`, so the line could be changed as follows
```suggestion
        except RuntimeError:
```

trz42 (changes requested):
Thanks for the update @laraPPr !

I added two small suggestions.

trz42 (approved):
Nice. Thanks @laraPPr !

PR #257, opened 2024-02-26
only respond to bot commands

trz42:
This let the bot only respond to comments which contain a bot command. If so it will either process the command if the sender is authorised to send commands to the bot or it will just respond with a note that the sender is not authorised.

The PR has been tested by letting two different accounts (one authorised, one not authorised) send comments with and without bot commands. See comment https://github.com/trz42/software-layer/pull/72#issuecomment-1965222781 and below.

The PR essentially adds a new function `contains_any_bot_command` and rearranges the processing in `handle_issue_comment_event` by first checking if the comment was newly created (only newly created comments are considered for processing bot commands), then checking if it contains any bot command, then proceeding with the processing of the commands (incl checking if the sender has permission to send a bot command).

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -31,7 +31,8 @@ import tasks.deploy as deploy
 from tasks.deploy import deploy_built_artefacts
 from tools import config
 from tools.args import event_handler_parse
-from tools.commands import EESSIBotCommand, EESSIBotCommandError, get_bot_command
+from tools.commands import EESSIBotCommand, EESSIBotCommandError,
+    contains_any_bot_command, get_bot_command
 from tools.permissions import check_command_permission
 from tools.pr_comments import create_comment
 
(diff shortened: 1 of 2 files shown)


eessi_bot_event_handler.py
```diff
@@ -113,7 +114,25 @@ def handle_issue_comment_event(self, event_info, log_file=None):
...
+            # NOTE we do not respond to an updated PR comment with yet another
+            #      new PR comment, because it would make the bot very noisy or
+            #      worse could result in letting the bot enter an endless loop
+            self.log(f"comment action '{action}' not handled")
+            return
+        # at this point we know that we are handling a new comment
+
+        # check if comment does not contain a bot command
+        if not contains_any_bot_command(comment_received):
+            self.log(f"comment does not contain a bot comment; not processing it further")
```
Neves-P, line 131: No need for an f string here, a plain string will do.
Can't claim credit, the CI clued me in for this one 😉

Neves-P (changes requested):
Looks great! Apart from the very small change the CI brings up I can't think of anything else to change. +1 for testing it "in the wild"

Neves-P (approved):

PR #266, opened 2024-04-22
reorganisation of constants and validation of config settings

trz42:
This PR turned out to be larger than originally intended (just validating that required config settings are defined when bot components start). However, it reorganises / brings consistent structure to how and where string constants are defined. Hopefully, this makes the code easier to read.

- string constants being used have been defined in the correct modules
  - when they are used they have to be prefixed with the module, hence it becomes more clear to which module they belong
- a new file tools/cvmfs_repository.py for constants used for the settings of a CernVM-FS repository is added
- both the event handler and job manager verify at the start that required/recommended settings are defined in 'app.cfg'

TODO
- [x] test that it actually works
  - various tests ran in https://github.com/trz42/software-layer/pull/74 
  - added issue https://github.com/EESSI/eessi-bot-software-layer/issues/267 for future improvements
- [x] fix potential hound issues

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -609,7 +609,11 @@ def main():
     opts = event_handler_parse()
 
     # config is read and checked for settings to raise an exception early when the event_handler starts.
-    config.check_required_cfg_settings(REQUIRED_CONFIG)
+    if config.check_required_cfg_settings(REQUIRED_CONFIG) == True:
+        print("Configuration check: PASSED")
+    else:
+        print("Configuration check: FAILED")
+        sys.exit(1)
     github.connect()
 
     if opts.file:
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -26,10 +26,8 @@ import waitress
 
 # Local application imports (anything from EESSI/eessi-bot-software-layer)
 from connections import github
-import tasks.build as build
 from tasks.build import check_build_permission, get_architecture_targets, get_repo_cfg, \
     request_bot_build_issue_comments, submit_build_jobs
-import tasks.deploy as deploy
 from tasks.deploy import deploy_built_artefacts
 from tools import config
 from tools.args import event_handler_parse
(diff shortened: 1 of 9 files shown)


tools/job_metadata.py
```diff
@@ -21,13 +21,53 @@
...
+JOB_PR_PR_COMMENT_ID = "pr_comment_id"
+
+# JWD/_bot_jobJOBID.result
+JOB_RESULT_SECTION = "RESULT"
 JOB_RESULT_ARTEFACTS = "artefacts"
+JOB_RESULT_COMMENT_DESCRIPTION = "comment_description"
 JOB_RESULT_FAILURE = "FAILURE"
-JOB_RESULT_SECTION = "RESULT"
 JOB_RESULT_STATUS = "status"
 JOB_RESULT_SUCCESS = "SUCCESS"
```
Neves-P, line 65: Should this not be lower case as it is part of the RESULT section?
```suggestion
JOB_RESULT_SUCCESS = "success"
```

tools/job_metadata.py
```diff
@@ -21,13 +21,53 @@
...
 JOB_PR_SECTION = "PR"
+JOB_PR_REPO = "repo"
+JOB_PR_PR_NUMBER = "pr_number"
+JOB_PR_PR_COMMENT_ID = "pr_comment_id"
+
+# JWD/_bot_jobJOBID.result
+JOB_RESULT_SECTION = "RESULT"
 JOB_RESULT_ARTEFACTS = "artefacts"
+JOB_RESULT_COMMENT_DESCRIPTION = "comment_description"
 JOB_RESULT_FAILURE = "FAILURE"
```
Neves-P, line 63: Should this not be lower case as it is part of the RESULT section?
```suggestion
JOB_RESULT_FAILURE = "failure"
```

Neves-P (changes requested):
Note: I have not run the bot instance on our site with this PR yet, I have only reviewed the code.

I think it's a good change and it adds needed structure for the bot configuration which will make it easier to setup on other sites too! I've just noted a few nitpicking things which could just be me not understanding the logic, go ahead and reject if that's the case.

tools/cvmfs_repository.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Local application imports (anything from EESSI/eessi-bot-software-layer)
+# (none yet)
+
+
+# access to a CernVM-FS repository is defined via a repos.cfg file and associated
+# tarballs containing configuration settings per repository
+# below, we define constants for the settings of each repository;
+# the section name, eg, 'eessi-2023.06' in '[eessi-2023.06]' is not fixed and
+# therefore no constant is defined for itsections and 'settings' in these files
```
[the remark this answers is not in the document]
trz42: You were right. It was not very clear, so I overhauled the whole comment in https://github.com/EESSI/eessi-bot-software-layer/pull/266/commits/1ceab6fe798f45546627d7c210ca2c90643c76b8

eessi_bot_event_handler.py
```diff
@@ -39,11 +37,59 @@
...
-REPO_TARGET_MAP = "repo_target_map"
+REQUIRED_CONFIG = {
+    config.SECTION_ARCHITECTURETARGETS: [
+        config.ARCHITECTURETARGETS_SETTING_ARCH_TARGET_MAP],       # required
+    config.SECTION_BOT_CONTROL: [
+        config.BOT_CONTROL_SETTING_COMMAND_PERMISSION,             # required
+        config.BOT_CONTROL_SETTING_COMMAND_RESPONSE_FMT],          # required
+    config.SECTION_BUILDENV: [
+        config.BUILDENV_SETTING_BUILD_JOB_SCRIPT,                  # required
+        config.BUILDENV_SETTING_BUILD_LOGS_DIR,                    # optional+recommended
```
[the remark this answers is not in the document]
trz42: That's a good suggestion. I added issue #268 for that.

tools/job_metadata.py
trz42: This is maybe a bit confusing, because some constants represent the names of settings (in a `_bot_jobJOBID.result` file): `JOB_RESULT_ARTEFACTS`, `JOB_RESULT_COMMENT_DESCRIPTION` and `JOB_RESULT_STATUS`.

Other constants (`JOB_RESULT_FAILURE` and `JOB_RESULT_SUCCESS`) represent values of the setting `JOB_RESULT_STATUS`. These values are determined by the `bot/check-build.sh` script and then written to a `_bot_jobJOBID.result` file.

I restructured the definition of the constants and added some comments to explain what they are used for.

See commit https://github.com/EESSI/eessi-bot-software-layer/commit/d3b211c405e1b4be64fc467457fb5425773393a4

tools/job_metadata.py
trz42: See comment above.

Neves-P (approved):
Looks great to me!

PR #273, opened 2024-06-21
add setting to give all jobs a unique name

trz42:
Until now all jobs are named after the job script. This PR adds a new configuration setting `job_name` that is used as the job's name. This is then also used by the job manager to list all jobs (with that name). Thus it is possible to have multiple bot instances running at the same cluster where each bot instance uses a different job name. Hence, each bot instance is able to only handle its "own" jobs. We can use this to run a separate bot instance, for example, for bot development and testing.


eessi_bot_job_manager.py
```diff
@@ -85,6 +87,10 @@ def __init__(self):
         cfg = config.read_config()
         job_manager_cfg = cfg[config.SECTION_JOB_MANAGER]
         self.logfile = job_manager_cfg.get(config.JOB_MANAGER_SETTING_LOG_PATH)
+        buildenv_cfg = cfg[config.SECTION_BUILDENV]
+        self.job_name = buildenv_cfg.get(config.BUILDENV_SETTING_JOB_NAME)
+        if len(self.job_name) < 3:
```
boegel, line 92: What if `job_name` is not set, this will go down pretty hard?

```python
>>> len(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type 'NoneType' has no len()
```

So:

```suggestion
        if self.job_name and len(self.job_name) < 3:
```

eessi_bot_job_manager.py
```diff
@@ -105,7 +111,7 @@ def get_current_jobs(self):
         if username is None:
             raise Exception("Unable to find username")
 
-        squeue_cmd = "%s --long --noheader --user=%s" % (self.poll_command, username)
+        squeue_cmd = "%s --long --noheader --user=%s --name='%s'" % (self.poll_command, username, self.job_name)
```
boegel, line 114:
```suggestion
        squeue_cmd = "%s --long --noheader --user=%s" % (self.poll_command, username)
        if self.job_name:
            squeue_cmd += "--name='%s'" % self.job_name

File: tasks/build.py [hunk 3]
```diff
@@ -654,6 +662,7 @@ def submit_job(job, cfg):
         build_env_cfg[config.BUILDENV_SETTING_SLURM_PARAMS],
         time_limit,
         job.slurm_opts,
+        f"--job-name='{job_name}'",
```
boegel, line 665: Also here, only use `--job-name` is the job_name is not `None`?

boegel (changes requested):

eessi_bot_job_manager.py
trz42: Ok, to be on the safe side.

tasks/build.py
trz42: Addressed in https://github.com/EESSI/eessi-bot-software-layer/pull/273/commits/c614a96f801ae2c79f9f1fa9cd5d3cdecbc156db

tasks/build.py
```diff
@@ -67,6 +67,10 @@ def get_build_env_cfg(cfg):
 
     buildenv = cfg[config.SECTION_BUILDENV]
 
+    job_name = buildenv.get(config.BUILDENV_SETTING_JOB_NAME)
+    log(f"{fn}(): job_name '{job_name}'")
+    config_data = {config.BUILDENV_SETTING_JOB_NAME: job_name}
```
boegel, line 72: This won't work, since `config_data` is overwritten below?

You should change the line 76 below to:

```python
config_data[config.BUILDENV_SETTING_JOBS_BASE_DIR] = jobs_base_dir
```

This also shows this wasn't (re)tested?

boegel (changes requested):

tasks/build.py
trz42: Yep. Silly me.

trz42: Fixed. Tested. Should work. 😬

boegel (approved):

PR #276, opened 2024-08-21
add filter for accelerators and pytests

trz42:
This is a first PR towards better support for building for GPUs. It includes the following changes/additions:

- a new filter component `accelerator` (that can be added to to `bot: build ...` commands)
- adjusted processing of filter component names so that components must be given with at least three characters
- improved definition and use of strings for raising exceptions if anything went wrong with providing filters
- fixed code for removal of filters (the code was neither used nor tested, thus we were not aware that it would not always work)
- added tests for the removal of filters and for the new filter component (`accelerators`)
- updated tests that were broken by code changes

### DIFF
--- a/tests/test_tools_filter.py
+++ b/tests/test_tools_filter.py
@@ -10,12 +10,18 @@
 #
 
 # Standard library imports
+import copy
 
 # Third party imports (anything installed into the local Python environment)
 import pytest
 
 # Local application imports (anything from EESSI/eessi-bot-software-layer)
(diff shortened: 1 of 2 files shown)


laraPPr:
@trz42 I have had a look and seems to look good. The only thing that is add in this pr is that you can pass `accelorator` in the bot command nothing more right? All the other stuff will than in an other PR right?

laraPPr (approved):
lgtm

PR #278, opened 2024-09-08
use github API for downloading the diff

smoors:
required for downloading from a private repo using a github token.


trz42 (approved):
Looks good. Thanks @smoors !

boegel:
@smoors I don't see how this helps for private repos, wasn't that the point of looking into this?

smoors:
> @smoors I don't see how this helps for private repos, wasn't that the point of looking into this?

both curl and git (via libcurl) support using github tokens stored in the ~/.netrc file.
however, that doesn't work to download the diff unless you use the github API.

PR #279, opened 2024-09-10
add docs on private repos

smoors:


trz42:
Nice!

Wonder how that works for GitLab. We can figure that out later.

PR #287, opened 2024-10-18
make submit options module name configurable

smoors:
we need this to let 2 bots on 2 different clusters (with different required submit options) listen on the same target branch.


smoors:
this is actually not needed, the `det_submit_opts()` function can dispatch depending on the cluster it is in.

Issue #290, opened 2024-12-01
add suppor for 'silent' mode


boegel:
By default, the current implementation of the bot is very chatty:

- It reports for duty in every PR being opened;
- It acknowledges receiving a command, even when that command is not relevant to that instance;
- It adds a comment if it takes any action in response to a command;

Now that we have various bots running, a lot of semi-useless comments are being created.
As long as the bots are active and listening, those comments are basically always ignored.

So I think we should add support to the bot for staying silent unless asked to reveal itself in PRs and not to reply to every command it receives. It should always post a comment when it actually acts on a command though, of course.

I would argue that this "silent mode" should be enabled by default even, and that bot instances should either be configured to be chatty, or that commands should enable "verbose", which could be useful for debugging purposes.

PR #292, opened 2025-01-22
Bot only reports moving to trash_bin when relevant

Neves-P:
In this PR, the bot will only write a comment when a PR is closed if any directories were actually moved to the `trash_bin`. Without this, bots that didn't build anything but are listening on PRs will still comment that an empty list was moved to the trash bin directory.

I have not yet tested this live on a bot build, but can do so on the dev.eessi.io bot

This PR is related to #290 but does fix it.

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -636,33 +636,37 @@ class EESSIBotSoftwareLayer(PyGHee):
         self.log(f"PR {pr.number}: determining directories to be moved to trash bin")
         job_dirs = determine_job_dirs(pr.number)
 
-        # 2) Get trash_bin_dir from configs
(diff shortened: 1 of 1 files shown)


boegel:
@Neves-P Looks look, but perhaps better test on a live bot to make 100% sure?

Neves-P:
@boegel , agreed! I'd like to try out in a live bot (but not the main bots). I should be able to try it on the dev.eessi.io bot, or make a new bot instance in RUG.

I hope to have a bit of time to try it this week, if not I will do so early next week 👍

trz42:
> @boegel , agreed! I'd like to try out in a live bot (but not the main bots). I should be able to try it on the dev.eessi.io bot, or make a new bot instance in RUG.
> 
> I hope to have a bit of time to try it this week, if not I will do so early next week 👍

Verified that the change works. See https://github.com/trz42/software-layer/pull/77

PR #293, opened 2025-01-29
Added information in README.md on how to create a ReFrame configuration file for the `software-layer` test step

casparvl:
Note that we can discuss if this is the right place for this documentation, since the bot is (in principle) generic. However, if we put this in `software-layer` itself, I'm afraid people deploying bot instances to build for the EESSI software layer will not find it. For now, I think this is the best place.

Imho, I'd think that _all_ documentation currently in the `README.md` should eventually move to the https://eessi.io/docs pages anyway for better visibility, but that's not urgent. In any case, that'll resolve the question of where to put this part that is slightly software-layer specific, as it _definitely_ is appropriate on https://eessi.io/docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -871,7 +871,7 @@ site_configuration = {
     ],
     'logging': common_logging_config(),
 }
-
+```
 
 ## Approach 2: describing a virtual node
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -765,7 +765,236 @@ moved_job_dirs_comment = PR merged! Moved `{job_dirs}` to `{trash_bin_dir}`
...
+            'name': 'default',
+            'cc': 'cc',
+            'cxx': '',
+            'ftn': ''
+            }
+        ],
+    'general': [
+        {
+            'purge_environment': True,
+            'resolve_module_conflicts': False,  # avoid loading the module before submitting the job
```
laraPPr, line 869: Autodetection is missing here I think?
```suggestion
            'resolve_module_conflicts': False,  # avoid loading the module before submitting the job
            'remote_detect': True,
```

README.md
```diff
@@ -765,7 +765,236 @@ moved_job_dirs_comment = PR merged! Moved `{job_dirs}` to `{trash_bin_dir}`
...
+        }
+    ],
+    'logging': common_logging_config(),
+}
+
```
laraPPr, line 874: The closing of this code block is missing

laraPPr (changes requested):

README.md
casparvl: Fixed

laraPPr (approved):
lgtm

PR #299, opened 2025-02-05
Set the local_tmp that is configured for a site as tmpdir

casparvl:
This is what I'm using for [this build](https://github.com/EESSI/software-layer/pull/903#issuecomment-2638547844), which was succesful.

It resolves the issue I had that a having a `TMPDIR` set on a system would cause `mktemp` in the `create_tarball.sh` to error out, because the `TMPDIR` wasn't available in the container. The fundamental issue is that `local_tmp` is not unique, so we don't want to bind-mount that directly into the container, as it might be shared between jobs. 

In this PR, we use the `local_tmp` that is configured and use `mktemp` to create a new directory within. That's what we will then use as the `TMPDIR`. This ensures that the `app.cfg` fully controls the `TMPDIR` used. Note that one can _still_ define `local_tmp = \$TMPDIR` if one wants to use the `TMPDIR` that would be set in a job, so there is no loss of generality here - just more control for whomever configures the bot.

We then set `STORAGE=${TMPDIR}` in the `bot/build.sh`, so that anything step (e.g. the `eessi-install-software.sh`) that uses `--storage` as argument to the `eessi_container.sh` will use this dir as a basis. Some of the steps will create subdirs within `TMPDIR`, e.g. with `mktemp`, or otherwise. That's all fine. Right now, they typically get bind-mounted as well, though strictly speaking, that'd no longer be needed since their parent dir (`$TMPDIR`) will be bind-mounted.

Note that putting this at the `bot-build.slurm` level is the only way in which we can globally override the `TMPDIR`. We could have tried to do this in each of the respective steps, but that means duplication of code, and if we ever add a step, we need to remember to do that as well. This is more robust.

### DIFF
--- a/scripts/bot-build.slurm
+++ b/scripts/bot-build.slurm
@@ -44,7 +44,7 @@ while IFS= read -r line; do
         continue
     fi
 
-    # If another section starts and we haven't found local_tmp, stop searching
+    # If another section starts and we haven't found local_tmp, don't try to match
     if [[ $line =~ ^\[.*\]$ && $inside_site_config == true ]]; then
         inside_site_config=false
     fi
--- a/scripts/bot-build.slurm
+++ b/scripts/bot-build.slurm
@@ -23,9 +23,6 @@
 #  - the directory may contain any additional files references in job.cfg,
 #    for example, repos.cfg and configuration file bundles for repositories
 
-# CVL EDIT 31-01-2025: set permissive umask for building
-umask 0002
-
 echo "Starting bot-build.slurm"
 EXPORT_VARS_SCRIPT=cfg/export_vars.sh
 if [ -f ${EXPORT_VARS_SCRIPT} ]; then
--- a/scripts/bot-build.slurm
+++ b/scripts/bot-build.slurm
@@ -23,6 +23,9 @@
 #  - the directory may contain any additional files references in job.cfg,
 #    for example, repos.cfg and configuration file bundles for repositories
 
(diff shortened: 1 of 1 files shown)


scripts/bot-build.slurm
```diff
@@ -32,6 +35,38 @@ if [ -f ${EXPORT_VARS_SCRIPT} ]; then
...
+while IFS= read -r line; do
+    # Check if we've reached [site_config]
+    if [[ $line =~ ^\[site_config\]$ ]]; then
+        inside_site_config=true
+        continue
+    fi
+
+    # If another section starts and we haven't found local_tmp, stop searching
+    if [[ $line =~ ^\[.*\]$ && $inside_site_config == true ]]; then
+        break
```
trz42, line 52: I'd suggest to change this line.
```suggestion
        inside_site_config=false
```
This would then also work in case the `job.cfg` would look like
```ini
[site_config]
some_key = some_value
[some_strange_empty_section]
[site_config]
local_tmp = i_have_no_reason_to_think_that_the_bot_would_create_such_a_config_file_but_who_knows_that_for_sure
```

trz42 (changes requested):
Tested in isolation. That is not with the bot, but just the code in a separate script and throwing different `job.cfg` files at it. Works as intended. I have a little suggestion, let me know if we should implement that or rather not. Happy to approve this after you let me know what you think.

trz42 (approved):
looks good to me

thanks @casparvl !

PR #300, opened 2025-02-06
add support for cloning Git repository via SSH rather than HTTPS

boegel:
This is useful when the target repository of PRs that the bot should act on is a private repository.

Opt-in via this in bot configuration file (`app.cfg`):
```ini
[buildenv]
clone_git_repo_via = 'ssh'
```

(draft PR since this needs actual testing, and perhaps more changes)


tasks/build.py
```diff
@@ -376,7 +382,16 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
     # - 'git checkout' base branch of pull request
     # - 'curl' diff for pull request
     # - 'git apply' diff file
-    clone_output, clone_error, clone_exit_code = clone_git_repo(f'https://github.com/{repo_name}', arch_job_dir)
+    log(f"Cloning Git repo via: {clone_via}")
+    if clone_via in (None, 'https'):
+        repo_url = f'https://github.com/{repo_name}'
```
smoors, line 387: nitpick, but github uses `.git` here too

tasks/build.py
```diff
@@ -44,6 +44,7 @@
 _ERROR_GIT_APPLY = "git apply"
 _ERROR_GIT_CHECKOUT = "git checkout"
 _ERROR_GIT_CLONE = "curl"
```
smoors, line 46: pre-existing, but shouldn't this be `git clone` instead of `curl`?

tasks/build.py
```diff
@@ -407,6 +422,17 @@ def download_pr(repo_name, branch_name, pr, arch_job_dir):
         error_stage = _ERROR_CURL
         return curl_output, curl_error, curl_exit_code, error_stage
 
+    git_diff_cmd = ' '.join([
+        f"git fetch origin pull/{pr.number}/head:{pr.number}",
+        f"git diff HEAD pr{pr.number} > {pr.number}.diff",
```
smoors, line 427: if you're now creating the like this, should'nt you remove the `curl_cmd` above?

tasks/build.py
boegel: Maybe, but the code wasn't doing that before, so I want to stick with what was used before.

But it works both with and without `.git`, so doesn't matter

tasks/build.py
boegel: Yeah, I guess so, I'll correct that after double checking

tasks/build.py
boegel: Yes, of course, I'll tackle that too

smoors: also, this is not exactly the same as the downloading the patch and applying it, as there can be changes pushed to the repo *after* the PR was created, which will be undone by this?

boegel:
This is working as expected now, so marking it ready-for-review.

Hopefully @trz42 has time to take a look?

tasks/build.py
boegel: @smoors made a good point (in Slack) about this change: this doesn't take into account that there may be changes in the `main` branch that are not included in the PR branch yet.

So, a better approach would be to:
* fetch the PR branch
* check out the PR branch
* `git merge main`
* swap back to `main` branch
* `git diff`
* `git apply`

trz42: This procedure hasn't been implemented yet, right?

A few thoughts:
- `main` should probably be the target branch of the PR, for example, for EESSI/2023.06 that would be 2023.06-software.eessi.io?
- the `git diff` after the swap back to "main" is really just the plain `git diff` or rather `git diff PR_branch`?

eessi_bot_event_handler.py
```diff
@@ -52,6 +52,7 @@
         config.BUILDENV_SETTING_BUILD_JOB_SCRIPT,                  # required
         config.BUILDENV_SETTING_BUILD_LOGS_DIR,                    # optional+recommended
         config.BUILDENV_SETTING_BUILD_PERMISSION,                  # optional+recommended
+        config.BUILDENV_SETTING_CLONE_GIT_REPO_VIA,                # optional
```
trz42, line 55: If not commented, the setting is 'required'.

```suggestion
        #config.BUILDENV_SETTING_CLONE_GIT_REPO_VIA,                # optional
```

README.md
```diff
@@ -443,6 +443,15 @@ variables) that are allowed to be specified in a PR command with the
...
 
+```
+clone_git_repo_via = https
+```
+
+The `clone_git_repo_via` setting specifies via which mechanism the Git repository
+should be cloned. This can be either:
+* `https` (default): clone repository via HTTPS with `git clone https://github.com/<owner>/<repo>`
+* `ssh`: clone repository via SSH with `git clone git@github.com:<owner>/<repo>.git`
+
```
trz42, line 454: Here's a suggestion about adding a little help for using 'ssh'.

````suggestion
In case of using 'ssh', one may need additional steps to ensure that the bot uses the right ssh key and does not ask for a passphrase (if the key used is protected with one). Here are a few things to consider:
- if the ssh key to be used does not have a standard name (e.g., `id_rsa`), add the following entry to `~/.ssh/config` in the bot's account
  ```
  Host github.com
    User git
    IdentityFile ~/.ssh/NAME_OF_PRIVATE_KEY_FILE
  ```
- if the key is protected by a passphrase (**highly recommended**), run an SSH agent and add the key to it
  ```
  eval $(ssh-agent -s)
  ssh-add ~/.ssh/NAME_OF_PRIVATE_KEY_FILE
  ```

````

File: app.cfg.example [hunk 6]
```diff
@@ -136,6 +136,10 @@ allow_update_submit_opts = false
 # exported into the build environment via `exportvariable` filters
 allowed_exportvars = ["NAME1=value_1a", "NAME1=value_1b", "NAME2=value_2"]
 
+# mechanisn to use to clone Git repository
+# 'https' to clone via HTTPS (git clone https://github.com/<owner>/<repo>)
+# 'ssh' to clone via SSH (git clone git@github.com:<owner>/<repo>.git)
```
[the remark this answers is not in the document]
trz42: Here's a suggestion about adding a little help for using 'ssh'.

````suggestion
# In case of using 'ssh', one may need additional steps to ensure that the bot
# uses the right ssh key and does not ask for a passphrase (if the key used is
# protected with one). Here are a few things to consider:
#  - if the ssh key to be used does not have a standard name (e.g., 'id_rsa'),
#    add the following entry to '~/.ssh/config' in the bot's account
#
#    Host github.com
#      User git
#      IdentityFile ~/.ssh/NAME_OF_PRIVATE_KEY_FILE
#
#  - if the key is protected by a passphrase (**highly recommended**), run an
#    SSH agent and add the key to it (with the following two commands)
#
#    eval $(ssh-agent -s)
#    ssh-add ~/.ssh/NAME_OF_PRIVATE_KEY_FILE

````

File: tasks/build.py (hunk 3 above)
Comment (boegel) [reply tasks/build.py:416]: I'll change this so the new approach based on `git` is only used when `clone_via` is set to `'ssh'` (so nothing changes for existing bots that listen to a public repo).

That gives us some freedom to experiment with the situation where the target branch has been updated since the branch for the PR was created.

File: eessi_bot_event_handler.py [hunk 7]
```diff
@@ -52,6 +52,7 @@
         config.BUILDENV_SETTING_BUILD_JOB_SCRIPT,                  # required
         config.BUILDENV_SETTING_BUILD_LOGS_DIR,                    # optional+recommended
         config.BUILDENV_SETTING_BUILD_PERMISSION,                  # optional+recommended
+        # config.BUILDENV_SETTING_CLONE_GIT_REPO_VIA,                # optional
         config.BUILDENV_SETTING_CONTAINER_CACHEDIR,                # optional+recommended
```
boegel, line 56:
```suggestion
        config.BUILDENV_SETTING_CONTAINER_CACHEDIR,                # optional+recommended
        # config.BUILDENV_SETTING_CLONE_GIT_REPO_VIA,                # optional

File: README.md [hunk 8]
```diff
@@ -443,6 +443,15 @@ variables) that are allowed to be specified in a PR command with the
...
 
 
+```
+clone_git_repo_via = https
+```
+
+The `clone_git_repo_via` setting specifies via which mechanism the Git repository
+should be cloned. This can be either:
+* `https` (default): clone repository via HTTPS with `git clone https://github.com/<owner>/<repo>`
+* `ssh`: clone repository via SSH with `git clone git@github.com:<owner>/<repo>.git`
```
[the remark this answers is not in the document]
boegel: fixed via 293ce3c5e3cff7c9f463aa4062b5576e5c1faab5

tasks/build.py
boegel: I've updated this in 80efece to only use the `git diff` approach when `clone_git_repo_via` is used.

It's still doing only `git fetch` + `git diff`, so you could still have issues when the PR branch is way behind on the target branch, I'll try to look into that next.

boegel:
I haven't actually re-tested this after the changes I've made, and some more changes are required to the `git diff` approach to take into account that the target branch of a PR may have been updated since the PR was opened, so I'm marking this as work-in-progress for now...

smoors:
this works as expected now that we're creating a diff from the `merge-base` instead of from the `HEAD`, which produces exactly the same diff as the one downloaded directly from api.github.com.

one thing that doesn't work with SSH is the `bot: status` command as that still requires using the github api. might be good to add a note about this.

tasks/build.py
```diff
@@ -481,6 +503,12 @@ def comment_download_pr(base_repo_name, pr, download_pr_exit_code, download_pr_e
             download_comment = (f"```{download_pr_error}```\n"
                                 f"{download_pr_comments_cfg[config.DOWNLOAD_PR_COMMENTS_SETTING_GIT_APPLY_FAILURE]}"
                                 f"\n{download_pr_comments_cfg[config.DOWNLOAD_PR_COMMENTS_SETTING_GIT_APPLY_TIP]}")
+        elif error_stage == _ERROR_PR_DIFF:
+            download_comment = (f"```{download_pr_error}```\n"
+                                f"{download_pr_comments_cfg[config.DOWNLOAD_PR_COMMENTS_SETTING_PR_DIFF_FAILURE]}"
+                                f"\n{download_pr_comments_cfg[config.DOWNLOAD_PR_COMMENTS_SETTING_PR_DIFF_TIP]}")
```
trz42, line 509: These settings should probably be added to the list of required settings.

trz42 (changes requested):
Minor comment.

Can you also pull in upstream changes, thus one could test this PR and the PR for allowing no spaces between `bot:` and `build`?

tasks/build.py
boegel: fixed in 750b8a4

PR #303, opened 2025-03-05
sign artefact and metadata file and upload signatures

trz42:
Adds signing of a tarball (artefact) and the corresponding metadata file and uploading those to the S3 bucket too.

Uses a sign script that could be provided by the target repository, for example, see https://github.com/EESSI/software-layer/pull/948

Because the latter uses the "recently" added option of `ssh-keygen` to sign files (option `-Y` requires OpenSSH v8.2 or newer), the bot may be configured to run the upload script in a container that provides a sufficiently recent version.

The PR adds the capability to skip the ReFrame tests at the end of a build job. This should not be used for production builds, but is useful to shorten development cycles.

The PR has been extensively tested via https://github.com/trz42/software-layer/pull/89 
- As part of the testing three bot instances were used: one on our main AWS build cluster, one on JSC and one on eX3.
- Where possible, signing was tested within and without a container.
- Eventually, all tarballs, metadata files and their signatures were successfully uploaded to an S3 bucket.
- A fix for a minor issue with the sign script (too open permissions for the converted key) has been suggested in https://github.com/EESSI/software-layer/pull/948


app.cfg.example
```diff
@@ -161,7 +161,11 @@ no_build_permission_comment = Label `bot:build` has been set by user `{build_lab
 allow_update_submit_opts = false
 
 # defines which name-value pairs (environment variables) are allowed to be
-# exported into the build environment via `exportvariable` filters
+# exported into the build environment via 'exportvariable' filters
+# The bot build script makes use of the variable 'SKIP_TESTS' to determine if
+# ReFrame tests shall be skipped or not. Default value is 'no'. If the value is
+# 'yes' and the exportvariable filter is added to a bot build command
+# ('export:SKIP_TESTS=yes'), ReFrame tests are skipped.
 allowed_exportvars = ["NAME1=value_1a", "NAME1=value_1b", "NAME2=value_2"]
```
Neves-P, line 169:
```suggestion
#allowed_exportvars = ["NAME1=value_1a", "NAME1=value_1b", "NAME2=value_2"]
```
This is not a change in this PR and a minor thing, but shouldn't `allowed_exportvars` be commented out by default in the example file (i.e., by default the dummy variables in the example aren't exported)

scripts/eessi-upload-to-staging
```diff
@@ -93,6 +96,13 @@ function display_help
   echo "                                    link the upload to a PR"               >&2
   echo "  -r | --repository FULL_NAME    -  a repository name ACCOUNT/REPONAME;"   >&2
   echo "                                    used to link the upload to a PR"       >&2
+  echo "  -s | --sign-script SCRIPT_PATH -  path to script that is used to sign"   >&2
+  echo "                                    artefacts and metadata files. The"     >&2
+  echo "                                    script is called with two arguments:"  >&2
+  echo "                                    KEY file_to_sign. The KEY is the one"  >&2
+  echo "                                    provided via option --sign-key. The"   >&2
+  echo "                                    latter is determined by this script."  >&2
+  echo "                                    [default: don't sign]"                 >&2
```
Neves-P, line 105:
```suggestion
  echo "                                    [optional; default: don't sign]"                 >&2
```

scripts/eessi-upload-to-staging
```diff
@@ -83,6 +83,9 @@ function display_help
   echo "                                    ingestion procedure"                   >&2
   echo "  -l | --list-variables          -  list variables that are available"     >&2
   echo "                                    for expansion"                         >&2
+  echo "  -k | --sign-key SCRIPT_KEY     -  specify location of the key to be"     >&2
+  echo "                                    used to sign artefacts and metadata"   >&2
+  echo "                                    files [default: don't sign]"           >&2
```
Neves-P, line 88:
```suggestion
  echo "                                    files [optional; default: don't sign]"           >&2
```

Neves-P (changes requested):
This looks great and is quite clear! I made just a few suggestions in the help messages and a default, but these are minor and I have no issues if you disagree and would rather merge as is.

Feel free to commit the suggestions or reject them and I'm happy to merge when done either way.

app.cfg.example
trz42: While it's only an example file, it's a good idea to comment this and add a note ala "This setting is optional. If you want to use it uncomment it and provide meaningful key-value pair(s)."

Neves-P (approved):
Looking good! :+1:

scripts/eessi-upload-to-staging
```diff
@@ -185,6 +213,21 @@ done
 # restore potentially parsed filename(s) into $*
 set -- "${POSITIONAL_ARGS[@]}"
 
+# ensure that either none or both of $sign_key and $sign_script are defined
+if [[ -n "${sign_key}" ]] && [[ -n "${sign_script}" ]]; then
+  sign=1
+elif [[ -n "${sign_key}" ]]; then
+  sign=0
```
boegel, line 220: Nitpicking, can maybe be tackled in a follow-up PR: I would've used `'no'` and `'yes'` here, to avoid confusion.

`0` is often a sign of success in bash scripts, here's it's a negative value, `'no'` leaves no room for confusion

PR #304, opened 2025-03-11
Add ssh signing script

ocaisa:


trz42 (approved):
Thanks a lot @ocaisa !

Script (signing part) has been tested extensively with the bot via https://github.com/trz42/software-layer/pull/89 Testing included the scenario that the script is provided as part of the bot (code).

The accompanying PR #303 includes necessary information about the configuration of the bot to use this script.

PR #307, opened 2025-03-14
avoid that host initialisation scripts are run when launching container

trz42:
Apparently launching a container may run some initialisation scripts from the host inside the container. This is likely not needed nor desired. This small PR adds the option `--contain` to when we launch a container to run the script that uploads files to an S3 bucket.


bedroge (approved):
Lgtm. Perhaps we should even do something similar for the build container?

PR #308, opened 2025-03-14
use bot name as namespace

trz42:
Use the bot instance's name as the namespace when signing files. Both sign and verify steps are updated. The verification is also prepared to support additional options such as `valid-before`.

CI tests are adjusted and extended to cover new functionality.

Part of CI output (added tests):
```
Verifying the signature against allowed signers...
Checking principal: wrong_namespace_identity and namespace: CI
Couldn't verify signature: namespace does not match
Could not verify signature.
Error: No valid signature found.
Expected failure for wrong namespace
Verifying the signature against allowed signers...
Checking principal: expired_key_identity and namespace: ci
allowed_signers:1: key has expired: verify time 2025-03-14T21:23:43 > valid-before 2025-03-11T00:00:00
Could not verify signature.
Error: No valid signature found.
Expected failure for expired key
```

### DIFF
--- a/scripts/eessi-upload-to-staging
+++ b/scripts/eessi-upload-to-staging
@@ -75,6 +75,8 @@ function display_help
   echo "                                    expansion will be applied; arg '-l'"   >&2
   echo "                                    lists variables that are defined at"   >&2
   echo "                                    the time of expansion"                 >&2
+  echo "  -b | --bot-instance NAME       -  name of the bot instance that uploads" >&2
+  echo "                                    files to S3"                           >&2
   echo "  -e | --endpoint-url URL        -  endpoint url (needed for non AWS S3)"  >&2
   echo "  -h | --help                    -  display this usage information"        >&2
   echo "  -i | --pr-comment-id           -  identifier of a PR comment; may be"    >&2
@@ -134,6 +136,7 @@ sign_key=
 sign_script=
 
 # provided via options in the bot's config file app.cfg and/or command line argument
+bot_instance=
 metadata_prefix=
 artefact_prefix=
 
@@ -145,6 +148,10 @@ while [[ $# -gt 0 ]]; do
   case $1 in
     -a|--artefact-prefix)
       artefact_prefix="$2"
+      shift 2
+      ;;
+    -b|--bot-instance)
+      bot_instance="$2"
       shift 2
       ;;
     -e|--endpoint-url)
@@ -263,7 +270,7 @@ for file in "$*"; do
         # 1st sign artefact, and upload signature
         if [[ "${sign}" = "1" ]]; then
           # sign artefact
-          ${sign_script} sign ${sign_key} ${file}
+          ${sign_script} sign ${sign_key} ${file} ${bot_instance}
           # TODO check if signing worked (just check exit code == 0)
           sig_file=${file}.sig
           aws_sig_file=${aws_file}.sig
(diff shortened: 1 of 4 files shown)


scripts/eessi-upload-to-staging
```diff
@@ -263,7 +270,7 @@ for file in "$*"; do
         # 1st sign artefact, and upload signature
         if [[ "${sign}" = "1" ]]; then
           # sign artefact
-          ${sign_script} sign ${sign_key} ${file}
+          ${sign_script} sign ${sign_key} ${file} ${bot_instance}
```
ocaisa, line 273: I think the arguments are now in a somewhat random order, which  is quite brittle. I will open a PR next week for the script to take proper arguments (`--namespace=<namespace>` etc.). That won't affect the format of the signatures though, so doesn't need to affect the ability to move forward with this PR and the general approach.

scripts/sign_verify_file_ssh.sh
```diff
@@ -102,19 +105,24 @@ if [ "$MODE" == "sign" ]; then
...
+        cat <<EOF
 
 For verification, your allowed signers file could contain:
-identity_1 $(cat "${TEMP_KEY}.pub")
+identity_1 namespaces="${NAMESPACE}",valid-before="LAST_VALID_DAY" $(cat "${TEMP_KEY}.pub")
 EOF
+    else
+       # refuse to sign without a namespace
+       echo "Error: missing namespace for signing"
+       exit $MISSING_NAMESPACE
```
ocaisa, line 118: Not sure why this needs to be an error, why not use a default like `file`? This is potentially not enforced by the allowed signers file anyway (namespace is an optional restriction).

I understand that _we_ want to leverage namespace, but other users of the bot may not. Again, I can implement in the follow-up PR.

scripts/sign_verify_file_ssh.sh
```diff
@@ -132,20 +140,16 @@ elif [ "$MODE" == "verify" ]; then
...
-            if ssh-keygen -Y verify -f "$ALLOWED_SIGNERS_FILE" -n file -I "$principal" -s "$SIG_FILE" < "$FILE_TO_SIGN"; then
-                echo "Signature is valid for principal: $principal"
-                exit 0
-            fi
-        done
+    while read -r principal options key
+    do
+        [[ -z "$principal" || "$principal" == \#* ]] && continue
+
+        namespaces=$(echo "$options" | grep -oP "namespaces=\"\K[^\"]+")
```
ocaisa, line 147: Are we introducing restrictions on how namespaces are defined here? Are `"` required by the format?

scripts/sign_verify_file_ssh.sh
```diff
@@ -132,20 +140,16 @@ elif [ "$MODE" == "verify" ]; then
...
-        IFS=',' read -ra principal_list <<< "$principals"
-
-        for principal in "${principal_list[@]}"; do
-            echo "Checking principal: $principal"
-            if ssh-keygen -Y verify -f "$ALLOWED_SIGNERS_FILE" -n file -I "$principal" -s "$SIG_FILE" < "$FILE_TO_SIGN"; then
-                echo "Signature is valid for principal: $principal"
-                exit 0
-            fi
-        done
+    while read -r principal options key
```
ocaisa, line 143: In general, full format support for the applied signers file is overly complicated for our use case (https://github.com/jelmer/dulwich/issues/1431#issuecomment-2492080239). I do think we need some kind of sanity check on the file (or lines in the file) to ensure we are getting the simplified format that we expect (single simple) for  principal and the optional namespace.

scripts/sign_verify_file_ssh.sh
trz42: Double-quotes around the values of options (such as `namespaces` or `valid-before`) are required.

scripts/sign_verify_file_ssh.sh
trz42: Maybe we can do that sanity check later?

ocaisa (approved):
LGTM, tests catch all the expected behaviours (at least for the EESSI use case)

Issue #294, opened 2025-01-31
Avoid assigning meaning to `arch_target_map` keys


casparvl:
Currently, the `arch_target_map` keys have meaning: they are interpreted by the `bot/build.sh` from `software-layer` to represent the OS/SUBDIR (but not accelerator). This is problematic, because if I have a system with e.g. `zen4` CPU nodes and `zen4+H100` GPU nodes, those would normally both be encoded as `linux/x86_64/amd/zen4` in the architecture map - and clearly I cannot do that because keys have to be unique.

It would be better if the keys were meaningless, and if the `bot/build.sh` would get it's information elsewhere. For example:
```
arch_target_map = {
    'virtual_partition_1': {
        'os': 'linux',
        'subdir': 'x86_64/amd/zen4',
        'slurm_params': '-p genoa <etc>',
    },
    'virtual_partition_2': {
        'os': 'linux',
        'subdir': 'x86_64/amd/zen4',
        'accel': 'nvidia/cc90',
        'slurm_params': '-p gpu_h100 <etc>',
    },
}
```
would then configure the cpu-only zen4 partition and zen4+H100 partition respectively.

This would require changes in two places:
1. The bot code, because the bot currently assumes that the _value_ of the `arch_target_map` is the slurm parameters to be used for submission. That should change, and it should extract one level deeper, i.e. `arch_target_map['some_partition']['slurm_params']` instead of `arch_target_map['some_os_subdir']`.
2. The `bot/build.sh` should extract the relevant information from a more deeply nested dict.

casparvl:
Changes will be needed _at least_ at around https://github.com/EESSI/eessi-bot-software-layer/blob/d4b4811664709546dff654e8c246949b6104346c/tasks/build.py#L603

**GOAL**

Suppose we have
```
arch_target_map = {
    # zen4 CPU nodes
    'virtual_partition_1': {
        'os': 'linux',
        'subdir': 'x86_64/amd/zen4',
        'slurm_params': '-p genoa <etc>',
    },
    # Zen4 CPU + H100 GPU nodes
    'virtual_partition_2': {
        'os': 'linux',
        'subdir': 'x86_64/amd/zen4',
        'accel': 'nvidia/cc90',
        'slurm_params': '-p gpu_h100 <etc>',
    },
    # Icelake + A100 GPU nodes
    'virtual_partition_3': {
        'os': 'linux',
        'subdir': 'x86_64/intel/icelake',
[...]

PR #312, opened 2025-04-17
Adapt arch target map

casparvl:
First changes - many more to come...


app.cfg.example
```diff
@@ -306,19 +306,78 @@ signing =
...
+    # This is a GPU partition. We specify an explicit "accel" property. Thus, only if the bot build command
+    # specifies that explicit accelerator in combination with the relevant CPU type,
+    # bot: build instance:xyz repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90
+    # will a build be triggered on this partition
+    # If you want to use this partition also for cross-compiling for cc70 and cc80 architectures, you can alter
+    # the "accel" property to 
+    # "accel": ["nvidia/cc70", "nvidia/cc80", "nvidia/cc90"]
+    # Note that setting:
+    # "accel": ["None", "nvidia/cc90"]
+    # is invalid here, since it would lead to both the cpu_zen4 and the gpu_h100 partitions matching the build command
```
trz42, line 368: "Invalid" in the sense the bot checks and refuses to even start or "invalid" in the sense it is not recommended because it's ambiguous?

trz42 (commented):
Left a bunch of comments.

This PR does actually more than what was originally "requested" in #294 

It provides more control for bot admins to allow/prevent certain jobs to be triggered, e.g., one might prevent accidental CPU-only builds on GPU-nodes. This feature is welcome, I think, although it changes the meaning of the `accelerator` filter. I don't know if the "prevent GPU builds on CPU-only nodes" is really necessary though. Main problem could be that is not so clear why some filter trigger a build and some other not.

While I'll have to give this a test run, I believe it does what it should do.

The move (or removal) of the `repo_target_map` is not necessary (I think, but I may be wrong). It doesn't hurt either.

I wonder if one should simply rename `arch_target_map` to make a braking change...

Also for the bot command filter names, I wonder if one could do better, e.g.,

| current    |   new |  comment |
|----------|----------------|--------------|
| `architecture` | `node` or `node_type` ... | `arch_target_map` could be renamed to `node_map` or `node_type_map` |
| | ... `cpu` | this would allow to make the definition of `node` types or partitions simpler _OR_ use the information for verifying (similar to what is done for `accel`) |

I wonder if we really need to have `os` and `cpu_subdir` in the definition of a partition, and use them to set the `architecture` (in the context when matching filters). Maybe it would be better to add another filter for `cpu` (defines for what CPU microarchitecture we want to build for). `os` and `cpu_subdir` are really odd things that should not matter much to the bot. They should be just passed through to the job.

Similarly, do we really care about the repository information when we want to trigger a job on a `zen4` node? It may add some options for an admin to constrain certain combinations ... however it feels a bit odd that this is a bot configuration option.

trz42:
There's actually one thing that may unnecessarily limit what we could achieve here. With the `slurm_params` we cannot only control to which type of node a job is submitted, we could also make use of partial nodes, different memory requirements, ...

For example,

```
"zen4_small_short" : {
  "slurm_params": "--partition zen4_nodes --ntasks-per-node=4 --mem=16G --time=0-6"
},
"zen4_big_long" : {
  "slurm_params": "--partition zen4_nodes --ntasks-per-node=8 --mem=64G --time=1-0"
}
```

but this would require a departure from the current filters (`architecture` and `accel`) to, maybe, (`nodetype`, `cpu` and `accel`).

casparvl:
> The move (or removal) of the repo_target_map is not necessary (I think, but I may be wrong). It doesn't hurt either.
> I wonder if one should simply rename arch_target_map to make a braking change...

Removing it keeps things clean - it is no longer used, as far as I could tell. But I do like your idea of renaming `arch_target_map`. And then in de config validation, maybe we can check if those items are still set - and if they are, print an informative message (that it's been replaced by `node_map` or whatever we will call it)

eessi_bot_event_handler.py
```diff
@@ -412,22 +411,22 @@ def handle_pull_request_opened_event(self, event_info, pr, req_chatlevel=ChatLev
...
-                            for repo_id in repo_ids]))
-        comment += "\n- repositories: "
-        if len(repositories) > 0:
-            comment += f"{', '.join([f'`{repo_id}`' for repo_id in repositories])}"
-        else:
-            comment += "none"
+        for partition_num, arch in enumerate(arch_map):
+            # Do not print virtual partition names, a bot admin may not want to share those
+            # Instead, just number them
+            comment += f"\n- Partition {partition_num+1}:"
```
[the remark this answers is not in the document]
casparvl: You don't need to now. They could be named `foo` and `bar`, it's not relevant to the end user (it's just helpful for the bot admin to pick a 'sensible' name :))

app.cfg.example
```diff
@@ -306,19 +306,78 @@ signing =
 
 [architecturetargets]
 # defines for which architectures the bot will build and what job submission
-#   parameters shall be used to allocate a compute node with the correct
+# parameters shall be used to allocate a compute node with the correct
+# The keys of the arch_target_map are virtual partition names. They don't have any meaning in the bot code,
+# and can thus be chosen as desired.
+# Note that you are responsible that ANY bot:build command ONLY matches a single virtual partition!
+# If multiple partitions match the same bot:build command, a failure will be triggered in the job dir preparation
```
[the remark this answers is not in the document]
casparvl: Well, it's not really a feature :P I think it is just something that emerges since you can now have multiple items in the `arch_target_map` that match a filter - that situation couldn't happen before I think. We need to consider how to deal with it. Currently, I give bot admins sufficient knobs to turn to make sure a filter only matches one partition. But we can deal with it differently if we so prefer - e.g. we could decide that the _first_ matching partition is the one where the job will be submitted, and we break the loop that matches the filter to partitions. In that case, the order in which partitions are listed in `arch_target_map` becomes important.

app.cfg.example
```diff
@@ -306,19 +306,78 @@ signing =
...
+    # This is a CPU partition. We specify an explicit "accel": "None" property. Thus, this partition will only be
+    # used if the bot build command does NOT contain an accel argument, e.g.
+    # bot: build instance:xyz repo:eessi.io-2023.06-software arch:zen4
+    # will cause the event_filter to mark this virtual partition as a valid match, and will use it to start building
+    # for zen4.
+    # When invoking the bot with an accelerator command, such as 
+    # bot: build instance:xyz repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90
+    # the event_filter will NOT mark this virtual partition as a valid match. This is intentional, as this particular
+    # (example) cluster has a native zen4+cc90 partition (gpu_h100) and we want this command to trigger a native build
+    # on that partition, rather than cross-compiling on this cpu_zen4 partition.
```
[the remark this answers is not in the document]
casparvl: It's always been a bit strange to me why `accel` had a different meaning than `arch`. I.e. `arch` was used for node selection, but `accel` wasn't. The only reason that 'worked' was because we always do native builds, and `arch` thus actually meant _both_: give me a node that matches this architecture _and_ build for this architecture (i.e. in this architectures prefix). Somehow, `accel` was special in that it _did not_ have any effect on node selection. But it _has to_ if we ever want to enable a combination of native and cross-compiled builds (how else do I ensure a native build?).

Anyway, I think this:

>  if zen4 architecture target defines "accel": ["None"] it is interpreted as a means to select or not select a build node. If cpu_zen4 architecture target does not define "accel" it is interpreted as a means to define for which accelerator we want to build for).

Should be seen differently. If a partition defines `"accel": ["None"]` that is a declaration that this partition is not suitable for (/should not be used for) compiling for accelerators. If a partition does not define `"accel"`, that is an (implicit) declaration that it may be used to (cross)-compile for _any_ accelerator.

The build command then defines what accelerator we want to build _for_ (e.g. it determines the prefix in which we'll install software). It is the job of the matching logic to then select a partition that can facilitate that request.

This is pretty equivalent to how ReFrame does things with partition `features`. A ReFrame config can declare "Partition A has a feature 'GPU', Partition B does not". A test can declare "I need a partition with the feature 'GPU' to run". As a result, when running ReFrame on a system with Partition A and B, that test will then only be scheduled on Partition A.

tasks/build.py
```diff
@@ -627,13 +610,34 @@ def prepare_jobs(pr, cfg, event_info, action_filter):
...
+    #     'virtual_partition_name2': {
+    # ... etc
+    for virtual_partition_name, partition_info in arch_map.items():
+        log(f"{fn}(): virtual_partition_name is {virtual_partition_name}, partition_info is {partition_info}")
+        # Unpack for convenience
+        arch_dir = partition_info['cpu_subdir']
+        if 'accel' in partition_info and accelerator is not None:
+            # Use the accelerator as defined by the action_filter. We check if this is valid for the current
+            # virtual partition later
+            arch_dir += accelerator
```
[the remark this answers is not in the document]
casparvl: You mean: the slash after zen4 is missing?

app.cfg.example
casparvl: Invalid in the sense it will lead to an error (the bot will fail on prepare the job dir for the gpu_h100 partition because that job dir was already prepared for the same job on cpu_zen4).

casparvl:
> However, it also feels like using the same hammer for another problem -- tweaking submission parameters -- maybe that would be better supported with another filter submitopts:arg=value,... and we do some sanity checking on the args and values (could do similar things what we do for exportvariable).

Hm, yeah, I'm not sure if 'the same hammer' is the solution here. Unless you consider 'max walltime' a requirement for the build job (just like a particular architecture or accelerator is). Then you could say, the person commanding the bot declares _what they need_, and they get a partition that has a max walltime that is _at least_ the amount they requested. Then, it would just be another filter option, essentially. And, we'd have to make that timelimit a separate field in the config (just like "os" and "cpu_subdir" are currently) so that it can easily be compared against. Finally, it would require then adding a `-t <that_config_field>` to the `slurm_params` at the end. Same could be true for memory btw.

app.cfg.example
trz42: Which would never happen if partitions have unique names and a user targets a partition by its name. Would also be easier to understand what is going on.

trz42:
> > I don't know if the "prevent GPU builds on CPU-only nodes" is really necessary though. Main problem could be that is not so clear why some filter trigger a build and some other not.
> 
> You don't _have_ to prevent GPU builds on CPU only nodes. But we _need_ the possibility to do this, because otherwise, two partitions match, and the bot will try to prep the job _twice in the same builddir_. See [this issue](https://github.com/EESSI/software-layer/pull/1120#issuecomment-3019629436)

To me this sounds the concept is not right if we need extra configuration to prevent something.

> The alternative way of fixing this issue is to break the loop after the first match. But in that case _the order of partitions_ in the `arch_target_list` would be important. I think as long as we document it well, that's also fine. But I think the same is true with the current setup: it requires proper docs, but then it offers the bot admin to do whatever they want :)

Strange argument. The current code uses the key "OS/CPU_ARCH" for defining part of the installation directory. That needs to be changed.

casparvl:
Discussed some things with Thomas on chat, some conclusions:

The real issue may not have been that keys have meaning, but that the `architecture` bot argument is used for two purposes: "Give me a node of CPU arch X", but also "Build for CPU arch X". For CPU, that was ok (at least for EESSI, since we didn't cross-compile). However, this PR now does the same to `accel`: it _used_ to only mean "Build for GPU arch Y", but now _also_ means "Give me a node with GPU arch Y".

The proposed solution is to split the two concerns by making two separate bot arguments, e.g.:

`bot:build on:CPU=x86_64/amd/zen4 for:CPU=x86_64/amd/zen4`

or 

`bot:build on:CPU=x86_64/amd/zen4,GPU=nvidia/cc90 for:CPU=x86_64/amd/zen4,GPU=nvidia/cc90`

The bot config could then be:

```
"cpu_zen4" : {
    "cpu_arch": "x86_64/amd/zen4",
    "slurm_params": "--partition genoa ...other-non-common-submission-params..."
}
"gpu_zen4_h100" : {
    "cpu_arch": "x86_64/amd/zen4",
    "gpu_arch": "nvidia/h200",
    "slurm_params": "--partition gpu_h100 ...other-non-common-submission-params..."
}

Some questions that were still open in the discussion:

- How do you make sure the `zen4` CPU-only builds don't end up on your h100 partition?
- How do you deal with requirements that match _multiple_ node types in the bot config?

The suggestion by @trz42 was to add `gpu_arch` to the context, and then **require that all elements of the context be present as a filter** (i.e. all these elements should have been passed as an argument for `bot:build`). This means that for the `cpu_zen4` node type, there is no `gpu_arch` in the context. Since `bot:build on:CPU=x86_64/amd/zen4 for:CPU=x86_64/amd/zen4` also doesn't set a filter for `gpu_arch`, this is a match. However, for the `gpu_zen4_h100` node type, the context _will_ have a `gpu_arch` set. So, if we enforce that all elements of a context need to be present as a filter, `bot:build on:CPU=x86_64/amd/zen4 for:CPU=x86_64/amd/zen4` doesn't match, becaues it doesn't set a `gpu_arch`. This _may_ require changes from PR310.

**Comment by casparvl** [comment]
Just a note: the `for` arguments should probably _not_ be stored as action filters. They are _meant_ to be used for selection, and thus should not be compared to the context. The only reason to put them as a filter would then be as a means to pass on the values from the bot command line, to eventually the `prepare_job_cfg`. That's a bad reason to do it like that. We should really make a separate place to store these from `EESSIBotCommand.__init__() (i.e. some separate attribute), then pass it on when `submit_build_jobs()` calls `prepare_jobs()` and then `prepare_job_cfg`

**Comment by casparvl** [comment]
Reporting of which arch it builds for: https://github.com/casparvl/software-layer/pull/1#issuecomment-3057507437

File: tasks/build.py [hunk 6]
```diff
@@ -823,7 +868,7 @@ def submit_job(job, cfg):
     # instances run on the same system
     job_name = cfg[config.SECTION_BUILDENV].get(config.BUILDENV_SETTING_JOB_NAME)
 
-    # add a default time limit of 24h to the job submit command if no other time
+    # add a default time limit of 24h to the job submit comnand if no other time
```
casparvl, line 871: Typo

app.cfg.example
```diff
@@ -356,12 +393,15 @@ scontrol_command = /usr/bin/scontrol
 #         are removed, the output (in PR comments) will lack important
 #         information.
 [submitted_job_comments]
-# awaits_release is no longer used since bot release v0.7.0
-# awaits_release = job id `{job_id}` awaits release by job manager
+awaits_release = job id `{job_id}` awaits release by job manager
```
bedroge, line 396: Why is the re-enabled? Based on that older comment, it shouldn't be used anymore?

eessi_bot_event_handler.py
```diff
@@ -411,23 +413,21 @@ def handle_pull_request_opened_event(self, event_info, pr, req_chatlevel=ChatLev
...
-        comment += "\n- repositories: "
-        if len(repositories) > 0:
-            comment += f"{', '.join([f'`{repo_id}`' for repo_id in repositories])}"
-        else:
-            comment += "none"
+        node_map = get_node_types(self.cfg)
+
+        comment = f"Instance `{app_name}` is configured to build on:"
+        for node in node_map:
+            comment += f"\n- Partition `{node}`:"
```
bedroge, line 420: Not sure if I would use partition here. If we do, maybe we should also use/recommend that in the app.cfg.example (instead of using build node name). But otherwise I'd just remove "Partition" here.

eessi_bot_event_handler.py
```diff
@@ -588,18 +588,23 @@ def handle_bot_command_status(self, event_info, bot_command):
...
             comment_status += f"{status_table['status'][x]}|"
             comment_status += f"{status_table['url'][x]}|"
 
         self.log(f"Overview of finished builds: comment '{comment_status}'")
         issue_comment = create_comment(repo_name, pr_number, comment_status, ChatLevels.MINIMAL)
-        return issue_comment
+        if issue_comment:
+            return f"\n  - added status comment {issue_comment.html_url}"
+        else:
+            return "\n  - failed to create status comment"
```
bedroge, line 607: Not sure why this used to return an IssueComment instance. Looking at the other bot command handlers, it makes sense to me that this also returns a string. Can you update the docstring of the function accordingly?

eessi_bot_event_handler.py
```diff
@@ -692,7 +697,7 @@ def main():
     opts = event_handler_parse()
 
     # config is read and checked for settings to raise an exception early when the event_handler starts.
-    if config.check_required_cfg_settings(REQUIRED_CONFIG):
+    if config.check_cfg_settings(REQUIRED_CONFIG):
```
bedroge, line 700: `REQUIRED_CONFIG` is defined in this file, but the new `FORBIDDEN_CONFIG` is added in `config.py`. Unless I'm overlooking a good reason to have them in different files, I'd say that they belong in the same file? And probably `config.py` makes more sense? And then the function can simply be called without arguments.

bedroge (changes requested):
Part 1 :wink:

tasks/build.py
```diff
@@ -925,30 +948,43 @@ def create_pr_comment(job, job_id, app_name, pr, symlink):
...
     """
     fn = sys._getframe().f_code.co_name
 
-    # obtain arch from job.arch_target which has the format OS/ARCH
-    arch_name = '-'.join(job.arch_target.split('/')[1:])
+    # Obtain the architecture on which we are building from job.arch_target, which has the format OS/ARCH
+    on_arch = '-'.join(job.arch_target.split('/')[1:])
+
+    # Obtain the architecture to build for
+    for_arch = build_params[BUILD_PARAM_ARCH]
```
bedroge, line 963: I started wondering here if build_params shouldn't be part of the job namedtuple, since they both contain information about the build job?

app.cfg.example
casparvl: You're right, this branch probably forked before that change

casparvl: Fixed

eessi_bot_event_handler.py
casparvl: Good point, done. It was definitely a wrong type before, it gave me a hard error.

tools/commands.py
```diff
@@ -85,19 +86,76 @@ def __init__(self, cmd_str):
...
+
+            # Add the other filter args to the normalized filters. The other_filter_args are already colon-separated
+            # so no special parsing needed there
+            log(f"Extracted filter arguments related to hardware target: {normalized_filters}")
+            log(f"Other extracted filter arguments: {other_filter_args}")
+            normalized_filters += other_filter_args
+
+            # Finally, change into a space-separated string, as expected by EESSIBotActionFilter
+            # e.g "arch:amd/zen2 accel:nvidia/cc90 repo:my.repo.io"
+            if normalized_filters:
```
bedroge, line 146: Not sure if this can be easily changed, but here I again got the feeling that it would be nice to have all this build information (with `on` and `for`, etc) in some combined object, which could do the parsing and store it in such a way that you can easily get the right information out of it. It now feels like we are doing a lot of juggling around with those arguments in different places of the code. Perhaps it's not realistic and I'm overlooking things, but just wanted to throw out the idea.

eessi_bot_event_handler.py
casparvl: The `event_handler` and `job_manager` each have a _different_ `REQUIRED_CONFIG`, which is why they are defined there instead of in a central place like `config.py`. The `FORBIDDEN_CONFIG` is central: it lists the items that should simply no longer be defined in the config since they are not used by _any script_ anymore.

tasks/build.py
```diff
@@ -627,13 +612,35 @@ def prepare_jobs(pr, cfg, event_info, action_filter):
...
     jobs = []
-    for arch, slurm_opt in arch_map.items():
-        arch_dir = arch.replace('/', '_')
-        # check if repo_target_map contains an entry for {arch}
-        if arch not in repocfg[config.REPO_TARGETS_SETTING_REPO_TARGET_MAP]:
-            log(f"{fn}(): skipping arch {arch} because repo target map does not define repositories to build for")
+    # This loop assumes the following structure for arch_target_map
+    # Note that 'accel' is a list, to easily allow a single CPU partition to be used for cross compilation
+    # for a lot of accelerator targets
+    # arch_target_map = {
```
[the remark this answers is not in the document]
casparvl: I'll remove this altogether. Thomas already commented that this should basically be made clear in the `app.cfg.example`, and indeed, that should then be the single source of truth for the description of the structure.

bedroge (changes requested):

tasks/build.py
casparvl: The build parameters (i.e. anything passed to `for:` on the command line) need to somehow make their way from the `bot_command` as it is present in `handle_bot_command_build` to this point (and to the `prepare_jobs()` function as well, where it is also needed). Short of passing the raw `bot_command` in it's entirety, the best way to do this is to create a separate data structure for it.

Note that `create_pr_comment` indeed receives _both_ `job` and `build_params`. But `prepare_jobs()` doesn't: instances of the `jobs` `NamedTuple` only get create _in_ that function. Thus, the only way to get things from `handle_bot_command_build` to `prepare_jobs()` is to pack them in a data structure, and pass them as arguments through the respective functions.

The only alternative I see is passing the full `bot_command` down to `prepare_jobs`, but I would consider that bad practice, as it creates less isolated code by passing more information than is required.

tasks/build.py
```diff
@@ -1085,11 +1132,66 @@ def check_build_permission(pr, event_info):
...
+    #    ("", None, None, None),
+    # ]
+    formatter = string.Formatter()
+    regex_parts = []
+
+    for literal_text, field_name, _, _ in formatter.parse(format_str):
+        # We use re.escape to escape any special characters in the literal_text, as we want to match those literally
+        regex_parts.append(re.escape(literal_text))
+        if field_name is not None:
+            # Create a non-greedy, named capture group. Note that the name itself as an f-string
```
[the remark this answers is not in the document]
casparvl: Neh, I meant the `{field_name}`. The correct name for this is a "format specifier" I believe. I'll change the comment.

eessi_bot_event_handler.py
bedroge: Okay, I overlooked that, that does actually make sense and is a valid reason to have them split.

tasks/build.py
```diff
@@ -1085,11 +1115,83 @@ def check_build_permission(pr, event_info):
...
+            # We match any character, but in a non-greedy way. Thus, as soon as it can match the next
+            # literal text section, it will - thus assuming that that's the end of the field
+            # We use .* to allow for empty fields (such as the optional accelerator fields)
+            regex_parts.append(f"(?P<{field_name}>.*?)")
+
+    # Finally, make sure we append a $ to the regex. This is necessary because of our non-greedy matching
+    # strategy. Otherwise, a formatting string that ends with a formatting item would only match the first letter
+    # of the field, because it doesn't find anything to match after (and it is non-greedy). With the $, it has
+    # something to match after the field, thus making sure it matches the whole field
+    # This does assume that the format_str in the string to be matched is indeed followed by and end-of-line character
```
bedroge, line 1169:
```suggestion
    # This does assume that the format_str in the string to be matched is indeed followed by an end-of-line character
```

bedroge (approved):

bedroge:
Tested in https://github.com/EESSI/eessi-bot-software-layer/pull/312/commits/2f3c0aee679a20f3f31c8e2c49cf587e4fc34ece.

PR #314, opened 2025-05-03
support different deployment tasks

trz42:
Extends the upload script to support different deployment tasks such as adding, deleting and updating contents of a CernVM-FS repository. For now, the script is called by the bot with default parameters corresponding to adding a tarball. This can then be used by the automated ingestion procedure on the Stratum-0 for processing different types of tasks.

In a subsequent PR, we will add support for different tasks created by job scripts. These could add files named
```
_bot_jobJOBID.task.SEQUENCE
```
The files would simply contain information such as
```
action = 'add' or 'delete' or 'update'
payload = /path/to/payload
order = some_32bit_number_used_for_ordering
mode = 'order' or 'timestamp'
```
The bot would do one upload for each of these task files.

### DIFF
--- a/scripts/eessi-upload-to-staging
+++ b/scripts/eessi-upload-to-staging
@@ -43,26 +43,107 @@ function create_metadata_file
   _repository=$3
   _pull_request_number=$4
   _pull_request_comment_id=$5
+  _action=$6
+  _order=$7
+  _mode=$8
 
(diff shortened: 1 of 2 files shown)


boegel:
@trz42 please sync with `develop` now that #316 is merged?

PR #324, opened 2025-06-19
make space before bot command optional

boegel:
`bot:build ...` or `bot: build` should both work just fine, no reason to require a hard space in between


boegel:
> Good idea 👍
> 
> Have you tested it?

I hadn't yet, and when I did I found a bug:
```
```

fixed in 66005da, also added log messages

trz42 (approved):
lgtm

PR #327, opened 2025-07-17
support for template values in `scontrol` command 

laraPPr:
This works but I'm not sure if the code now the code seems to specified to our systems. But it works.

Should still add some debugging if this seems like a good way to go forward. becuase now the but crashes if it has an unknow placeholder in the string that is set in `app.cfg`.

### DIFF
--- a/eessi_bot_job_manager.py
+++ b/eessi_bot_job_manager.py
@@ -118,6 +118,8 @@ class EESSIBotSoftwareLayerJobManager:
         squeue_cmd = "%s --long --noheader --user=%s" % (self.poll_command, username)
         if self.job_name:
             squeue_cmd += " --name='%s'" % self.job_name
+        # Format the output of SLURM
+        squeue_cmd += " --Format JobId:100@,Cluster:100@,Partition:100@,State:100@,Reason:100"
         squeue_output, squeue_err, squeue_exitcode = run_cmd(
             squeue_cmd,
             "get_current_jobs(): squeue command",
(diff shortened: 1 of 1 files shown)


boegel (changes requested):

eessi_bot_job_manager.py
```diff
@@ -296,6 +303,14 @@ def process_new_job(self, new_job):
         """
         job_id = new_job["jobid"]
 
+        # if placeholder is used in scontrol command
```
trz42, line 306: The code below is always executed, hence the comment could be made a little more specific. For example, 
```suggestion
        # processing placeholders in scontrol command which is defined in the bot's app.cfg (setting `scontrol_command`)
```

trz42 (changes requested):
Looks good to me.

Added a small suggestion regarding a comment. Great if you could update `app.cfg.example` and `README.md`

eessi_bot_job_manager.py
```diff
@@ -296,6 +303,14 @@ def process_new_job(self, new_job):
         """
         job_id = new_job["jobid"]
 
+        # processing placeholders in scontrol command which is defined in the bot's app.cfg (setting `scontrol_command`)
```
casparvl, line 306: Please give an example of what might be in the `scontrol_command` that requires replacing. And what would be a typical value of `new_job` so that this replacement would work. Maybe also add in which scenario this is useful - at first sight, one would just expect this to be the path to your site's `scontrol` command, why would that need placeholders?

PR #331, opened 2025-08-07
Make sure to print an informative message to the log when a `for:` argument is missing

### LINKED ISSUE #330: Print a nice error if `for:` is not used in a bot build command
@laraPPr just ran into this error:
```
AttributeError: 'EESSIBotCommand' object has no attribute 'build_params'
```

That's caused by not using `for:` in the build command, which is now required. If it's not used, `self.build_params` is not defined here: https://github.com/EESSI/eessi-bot-software-layer/blob/develop/tools/commands.py#L110.

We need to add a proper check for this, and throw a nice error if the build command is incorrect.

casparvl:
Set a default `self.build_params = None` on constructiong the `EESSIBotCommand`. Then, in handling the bot build command, don't submit a job if `self.build_params` wasn't defined

fixes #330

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -530,6 +530,12 @@ class EESSIBotSoftwareLayer(PyGHee):
         pr_number = event_info['raw_request_body']['issue']['number']
         pr = gh.get_repo(repo_name).get_pull(pr_number)
         build_msg = ''
+        # Require that build_params is defined, it is required. Otherwise, return early
+        if bot_command.build_params is None:
+            build_msg = "No 'for:' argument was passed to the bot:build command. This argumen is required, so "
+            build_msg += "not submitting build jobs"
+            return build_msg
+
         if check_build_permission(pr, event_info):
             # use filter from command
             submitted_jobs = submit_build_jobs(pr, event_info, bot_command.action_filters, bot_command.build_params)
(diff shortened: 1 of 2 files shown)


boegel:
@casparvl Correct me if I'm wrong, but this will only result in a log message, but nothing will be exposed to whoever issues this bot build command, correct?
I think we want to avoid giving the "user" the silent treatment here, since requiring a `for:` argument is a "breaking change" in how the bot should be used, so we should be vocal about it.

casparvl:
I was afraid you'd say that. Although it has to be said that this is definitely not the only mistake the user can make in order to get the silent treatment: if you provide arguments for which none of the contexts match, nothing will happen (and nothing will be reported either).

I know it's a bit different here since it's a _change_ compared to previous command syntax. But some problems are the same. For example: if the user doesn't supply the `instance` arguments, do you really want all bots telling you that you're lacking the `for:` argument?

If you still prefer that, I can look into it (I don't immediately know how to write an issue comment from this somewhat arbitrary part of the code - but I'm sure I can figure it out). But since the group of people that can command the bot (at least for EESSI) is limited to two hands full, it seems easier to me to just inform everyone very clearly.

As a side note: once we finalize the governance, I think we should start a (private) #builders and #deployers channel, corresponding to the rules described in the governance. This would provide an easy way to inform everyone for whom it is relevant of such changes :)

casparvl:
As discussed in the meeting on Tuesday, I think we should just merge this - it's a temporary problem until people are all familiar with the new syntax anyway. If it proves problematic down the line, we can try to improve in a subsequent PR.

boegel (approved):

Issue #332, opened 2025-08-08
Update the 'status' feature to reflect e.g. last build only


casparvl:
In a PR with builds for a lot of architectures, it can be tricky to figure out if everything has been built for. The `bot:status` command gives a nice overview, but includes failures. In the end, one is interested in having successful builds for a certain set of architectures.

It would be convenient if we can adapt the `bot:status` command to only show the last build result. E.g. `bot:status mode:all` (or just `bot:status`) for the traditional behaviour, and then or `bot:status mode:last` to get an overview of the build status of the last builds done for every architecture. This will make it easier to see for a reviewer if successful builds are present for all the officially supported architectures - and if not, which ones are missing.

PR #334, opened 2025-08-19
Add option to bot:status to print the status only for the last build

casparvl:
This PR implements functionality so that one can do:

```
bot:status last_build
```

Which will create a status table only contain the last build for each `for` architecture. This makes it _much_ easier to check if all architectures have been (successfully) build for in a given PR. That's almost essential for GPU builds, where the number of target architectures is currently 13 (CPU) * 3 (GPU) = 39 architectures.

[This example](https://github.com/casparvl/software-layer-scripts/pull/2#issuecomment-3201357360) was created with a bot running from [this commit](https://github.com/EESSI/eessi-bot-software-layer/pull/334/commits/7b284ba8d83688d4104b880f5c3484d0bb87c0b0)

Fixes https://github.com/EESSI/eessi-bot-software-layer/issues/332

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -604,7 +604,7 @@ class EESSIBotSoftwareLayer(PyGHee):
             status_table['timestamp'] = timestamps
 
             # Figure out the sorting indices, so that things are sorted first by the 'for arch', and then by 'date'
-            sorted_indices = sorted(range(len(status_table['for arch'])), key=lambda x: (status_table['for arch'[x], status_table['timestamp'][x]))
+            sorted_indices = sorted(range(len(status_table['for arch'])), key=lambda x: (status_table['for arch'][x], status_table['timestamp'][x]))
             # Reverse, so that the newest builds are first
             sorted_indices.reverse()
             # Apply the sorted indices to get a sorted table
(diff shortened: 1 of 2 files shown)


trz42 (commented):
Two comments before having looked into this in detail:
- Since we now use multiple bot instances it might be good to include the instance name in the overview table.
- If a build succeeded for an architecture, then a follow-up failed, the table would show the last build as failed while setting the `bot:deploy` label would lead to the last successful one being uploaded. Not sure if that would happen often and how to approach that issue.

casparvl:
Hm, initially I wanted to mimic as closely as possible what the bot does for the 'deploy last' policy. I concluded that was not possible, because the status command _solely_ relies on issue comments, and not on any variables that are in memory (which the deployment policy does, it checks state in the job dir etc). But I now realize I don't need that: the status _is_ one of the items in the table, so I have that information. Should not be too hard to modify the behavior to pick the last success, rather than the last build. That mimics the deployment behavior more closely, which I agree is preferable.

Regarding the instance name: should be doable: it's in the same issue comment as the rest of the build info, so should be trivial to capture.

trz42:
As discussed online, we should get the PR merged as is and postpone potential improvements to a subsequent release.

PR #335, opened 2025-08-20
don't replace self.scontrol_command in process_new_job

boegel:
required bug fix for enhancement implemented in #327, to keep placeholder and allow processing jobs that require different placeholder values

Without this, the first job that's processed determines the value for a placeholder like `%(cluster)s`, but multiple jobs being processed/released in sequence may belong to different clusters


laraPPr (approved):
lgtm

Issue #343, opened 2025-08-26
bot:status last_build throwing an index out of range


casparvl:
This https://github.com/EESSI/software-layer-scripts/pull/73#issuecomment-3223526532

Triggered:
```
[20250826-T12:10:37] WARNING: A crash occurred!
Traceback (most recent call last):
  File "/var/tmp/eessibot/venv/lib/python3.11/site-packages/pyghee/lib.py", line 194, in process_event
    self.handle_event(event_info, log_file=log_file)
  File "/var/tmp/eessibot/venv/lib/python3.11/site-packages/pyghee/lib.py", line 124, in handle_event
    handler(event_info, log_file=log_file)
  File "/var/tmp/eessibot/eessi-bot-software-layer/eessi_bot_event_handler.py", line 310, in handle_issue_comment_event
    update = self.handle_bot_command(event_info, cmd)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/tmp/eessibot/eessi-bot-software-layer/eessi_bot_event_handler.py", line 489, in handle_bot_command
    return handler(event_info, bot_command)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/tmp/eessibot/eessi-bot-software-layer/eessi_bot_event_handler.py", line 613, in handle_bot_command_status
    sorted_indices = sorted(
                     ^^^^^^^
  File "/var/tmp/eessibot/eessi-bot-software-layer/eessi_bot_event_handler.py", line 615, in <lambda>
    key=lambda x: (status_table['for arch'][x], status_table['timestamp'][x])
                                                ~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```
I'm not sure why, this seems to have worked in my tests, but... something to double check

casparvl:
Strange thing is, it succeeded after a few more builds...

https://github.com/EESSI/software-layer-scripts/pull/73#issuecomment-3223957928

casparvl:
Ah! I had it once more:

```
[20250826-T17:38:32] WARNING: A crash occurred!
Traceback (most recent call last):
  File "/gpfs/home4/casparl/EESSI/bot-instance/eessi_bot_casparvl_venv/lib64/python3.9/site-packages/pyghee/lib.py", line 170, in process_event
    self.handle_event(event_info, log_file=log_file)
[...]

PR #345, opened 2025-08-26
Make sure that builds for different repos don't get deduplicated by the last_build option

### LINKED ISSUE #344: bot:status last_build doesn't consider tarballs for different EESSI versions to be unique
https://github.com/EESSI/software-layer-scripts/pull/73#issuecomment-3224004226 prints a list of 14 items, while 28 builds where done (14 for 2023.06, and 14 for 2025.06). The issue is that it deduplicates e.g. zen2+2023.06 and zen2+2025.06, while it should consider those items as unique.

The fix should be on [this line](https://github.com/EESSI/eessi-bot-software-layer/blob/764c5c96a49c8d5cbf8b2068ef92a0972eb15ed7/eessi_bot_event_handler.py#L628) where it should not only check if the current 'for' target is already there, but also if, for that target, the EESSI version is the same.

casparvl:
Fixes https://github.com/EESSI/eessi-bot-software-layer/issues/344

Note that I also added a `self.log`, to log the status table before we start modifying for the `last_build` option. Reason is that I was unable to reproduce https://github.com/EESSI/eessi-bot-software-layer/issues/343 so the only chance we have at resolving that _if_ it happens again is to have some better logging in place and see if we spot anything strange.

### DIFF
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -591,6 +591,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         repo_name = event_info['raw_request_body']['repository']['full_name']
         pr_number = event_info['raw_request_body']['issue']['number']
         status_table = request_bot_build_issue_comments(repo_name, pr_number)
+        self.log(f"Retrieved status table from issue comments: {status_table}")
 
         if 'last_build' in bot_command.general_args:
             # If the bot command is something like 'bot:status =last_build', then only retain the last build for each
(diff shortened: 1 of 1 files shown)


casparvl:
Fix is tested in https://github.com/casparvl/software-layer/pull/4

A run from the current `develop` branch of the bot: https://github.com/casparvl/software-layer/pull/4#issuecomment-3224608239
A run from the current feature branch for this PR: https://github.com/casparvl/software-layer/pull/4#issuecomment-3224603202
A run from the current feature branch that shows that the last build is correctly taken if builds are identical (i.e. same 'for arch' and same 'repo'): https://github.com/casparvl/software-layer/pull/4#issuecomment-3224737763

Seems like it does what it should be doing.

eessi_bot_event_handler.py
```diff
@@ -625,7 +626,17 @@ def handle_bot_command_status(self, event_info, bot_command):
                 'on arch': [], 'for arch': [], 'for repo': [], 'date': [], 'status': [], 'url': [], 'result': []
             }
             for x in range(0, len(sorted_table['date'])):
-                if sorted_table['for arch'][x] not in status_table_last['for arch']:
+                # Check if the current 'for arch' AND repo are already in the status_table_last. If not, add it
```
bedroge:
```suggestion
                # Check if the current 'for arch' AND 'for repo' are already in the status_table_last. If not, add it
```

bedroge (approved):

Issue #342, opened 2025-08-22
pinned-dependencies check failing for MarkDown linting and smee Dockerfile


boegel:
for `containers/Dockerfile.smee-client`:
* [x] https://github.com/EESSI/eessi-bot-software-layer/security/code-scanning/8
  * should be fixed by #347 
* [x] https://github.com/EESSI/eessi-bot-software-layer/security/code-scanning/9
  * should be fixed by #351
  
for `.github/workflows/markdown-lint.yml`:
* [x] https://github.com/EESSI/eessi-bot-software-layer/security/code-scanning/10
  * should be fixed by #347

trz42:
Should all be fine now.

PR #347, opened 2025-11-06
fix dependency versions for npm and Docker

trz42:
fixes two issues in #342. Pinning of npm install when building the smee-client container did not work out as hoped for.


ocaisa (approved):
If CI is passing seems good to me

PR #352, opened 2025-11-12
fix handling of artefact name

trz42:
Since build jobs can create tarballs using other compression tools such as `zstd` which result in different suffixes than `.gz` the bot's deploy code does not work correctly. Initially all information is correctly taken from a job's result file (named `_bot_jobJOBID.result`), but in subsequent processing part of the information is dropped and a hardcoded suffix (`.tar.gz`) is used.

This PR just uses the full name of the artefact as provided by the `*.result` file.

Note, this is a hot fix. There may be other issues with the processing of artefact names (e.g., if the suffix is just `.tar` without any `.gz` or `.zstd`). Also, the timestamp that was previously used to recreate the full name of the tarball to be uploaded is not used any longer it seems (maybe the hound will complain).

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -236,14 +236,14 @@ def append_artefact_to_upload_log(artefact, job_dir):
         upload_log.write(f"{job_plus_artefact}\n")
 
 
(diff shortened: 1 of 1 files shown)


casparvl (approved):
Lgtm

PR #353, opened 2025-11-12
fix artefact filename parsing and cleanup unneeded information

trz42:
- fixes parsing for artefact filename (was too rigid and only working with `.tar.gz` suffix)
- cleans up variables not needed any longer (timestamp in some function)
- adds all components of artefact filename to data structure, so just in case we need it at some point we have the information now

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -533,11 +533,13 @@ def parse_artefact_filename(filename):
 
     match = re.match(pattern, filename)
     if match:
+        log(f"{funcname}(): parsing '{filename}' matched")
         prefix = match.group(1)
         timestamp = match.group(2)
         suffix = match.group(3)
         return prefix, timestamp, suffix
 
+    log(f"{funcname}(): parsing '{filename}' did NOT match")
     return None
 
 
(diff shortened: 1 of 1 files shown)


tasks/deploy.py
```diff
@@ -517,6 +516,33 @@ def determine_successful_jobs(job_dirs):
...
+    Args:
+        filename (string): name of artefact with format A-B-C-...-N-TIMESTAMP.tar(.gz|.zstd|...)
+
+    Returns:
+        tuple: (prefix, timestamp, suffix) or None if pattern doesn't match
+    """
+    funcname = sys._getframe().f_code.co_name
+
+    # capture everything before the last dash, then unix timestamp, then suffix
+    pattern = r'^(.+)-(\d{10,})(\..+)$'
```
bedroge, line 532: This will stop working on Sat Nov 20 2286 :stuck_out_tongue_winking_eye:

boegel: I say we fix it then...

bedroge: I'll add a reminder to my calendar :grin:

bedroge (approved):

PR #357, opened 2025-12-02
use requests library instead of 'curl' and obtain all comments

### LINKED ISSUE #356: Overview table from `bot:status last_build` of the bot is incomplete for large numbers of builds
See https://github.com/EESSI/software-layer/pull/1278#issuecomment-3601273996

It may be that we're exceeding some length when gathering the input data _or_ when producing the overview table?

trz42:
fixes #356 

PR looks much bigger than it actually is. For most lines, only indentation has been reduced by one level / 4 spaces (previous version: lines 1217 - 1356, new version: lines 1229 - 1368)

Other changes are:
- imports `requests` library
- replaces the existing logic relying on the `curl` command by using the `requests` library plus using the `Link` header in the response for obtaining the next _page_ of the paginated response (if PR has more than 100 comments) and handles an exception if any happens while querying GitHub's REST API
- variable that contains all comments changed from `comments` to `all_comments`
- old lines 1358-1359 are not needed any longer


casparvl:
Tested in https://github.com/casparvl/software-layer/pull/6, I can confirm it fixes the issue since this report https://github.com/casparvl/software-layer/pull/6#issuecomment-3689727282 is complete (even though there is >> 100 comments and the build is comment 130-or-so).

I'll still need to review the actual code, but this is promising at least :)

tasks/build.py
```diff
@@ -1207,154 +1208,163 @@ def request_bot_build_issue_comments(repo_name, pr_number):
...
-                on_arch_fmt_with_accel = on_arch_fmt.format_map(PartialFormatDict(on_accelerator=accelerator_fmt))
-                on_arch_re_with_accel = template_to_regex(on_arch_fmt_with_accel)
-                on_arch_match = re.match(on_arch_re_with_accel, comment_body[1])
+
+    url = f'https://api.github.com/repos/{repo_name}/issues/{pr_number}/comments'
+    all_comments = []
+
+    try:
+        while url:
+            response = requests.get(url, params={'per_page': 100})
```
casparvl: You should specify a `timeout` argument - default is 'None', which means it'll wait forever if it doesn't get a response. I'm not sure what a good value is, but I guess 10 would be pretty reasonable.

```suggestion
            response = requests.get(url, params={'per_page': 100}, timeout=10)
```

casparvl (changes requested):
Should specify a timeout to avoid indefinite hangs.

We should also realize that we might hit the unauthenticated rate limit (60 requests/hour) quite easily. A token could help, but that would probably have to be set in the config of the bot... Some dummy code of how it would be used here is then:

```
    headers = {}
    if token:
        headers["Authorization"] = f"Bearer {token}"
...
    response = requests.get(url, params={'per_page': 100}, headers=headers, timeout=10)
```

It is a bit weird, since this means the bot would need someone's token (i.e. identify as an individual)... but it could be an optional configuration. Having the capability at least means an easy fix (update of a bot config, rather than a full new bot code) if we ever hit this. I leave it up to you if you think it's worth implementing in this PR - I'm also fine in doing it in a follow-up PR, or even just postponing until it is _actually_ an issue.

Another thing that could be considered is to wrap this in a retry loop, to make it more resilient against transient 5xx errors or network hiccups.

Again, pseudo (well: unchecked) code:

```
while url:
        for attempt in range(1, max_retries + 1):
            try:
                resp = requests.get(
                    url,
                    params={"per_page": per_page},
                    headers=headers,
                    timeout=10,                 # avoid hanging forever
                )
                resp.raise_for_status()       # 4xx/5xx → exception
                break                         # success → exit retry loop
            except requests.HTTPError as http_err:
                # 5xx are usually transient – retry; 4xx are not
                if 500 <= resp.status_code < 600 and attempt < max_retries:
                    wait = 2 ** attempt
                    log.warning(
                        "Transient HTTP %s, retry %d/%d after %ds",
                        resp.status_code, attempt, max_retries, wait,
                    )
                    time.sleep(wait)
                    continue
                # re‑raise – caller will see the exception
                raise
            except requests.RequestException as req_err:
                # network glitches, timeouts, DNS failures, etc.
                if attempt < max_retries:
                    wait = 2 ** attempt
                    log.warning("Network error: %s – retry %d/%d after %ds",
                                req_err, attempt, max_retries, wait)
                    time.sleep(wait)
                    continue
                raise
```

Again, I'm not convinced we should do this in the current PR - we could also retry by just calling `bot:status` again. To me, a quick merge is more important, as it is currently very hard to get (complete) status overviews for builds with many targets (i.e. CPU+GPU), which is _exactly_ where this functionality is most essential. It's almost impossible to check if all CPU+GPU combinations for 2025.06 are covered, currently.

trz42:
> Should specify a timeout to avoid indefinite hangs.

Good point. Should be configurable, thus easy to change.

> We should also realize that we might hit the unauthenticated rate limit (60 requests/hour) quite easily. A token could help, but that would probably have to be set in the config of the bot... Some dummy code of how it would be used here is then:
> 
> ```
>     headers = {}
>     if token:
>         headers["Authorization"] = f"Bearer {token}"
> ...
>     response = requests.get(url, params={'per_page': 100}, headers=headers, timeout=10)
> ```

As far as I know, the bot already creates short-lived tokens (using its certificate). I guess that could be used here too.

> It is a bit weird, since this means the bot would need someone's token (i.e. identify as an individual)... but it could be an optional configuration. Having the capability at least means an easy fix (update of a bot config, rather than a full new bot code) if we ever hit this. I leave it up to you if you think it's worth implementing in this PR - I'm also fine in doing it in a follow-up PR, or even just postponing until it is _actually_ an issue.

I think there is no need for a 'personal' token here, and it wouldn't be a good solution either. No need to have it as an extra configuration setting.

> Another thing that could be considered is to wrap this in a retry loop, to make it more resilient against transient 5xx errors or network hiccups.

Generally, good idea to improve resilience, but I would consider this out-of-scope for this PR. Perfectly fine to open a new issue on that (if no issue on that exists yet).

> Again, pseudo (well: unchecked) code:
> 
> ```
> while url:
>         for attempt in range(1, max_retries + 1):
>             try:
>                 resp = requests.get(
>                     url,
>                     params={"per_page": per_page},
>                     headers=headers,
>                     timeout=10,                 # avoid hanging forever
>                 )
>                 resp.raise_for_status()       # 4xx/5xx → exception
>                 break                         # success → exit retry loop
>             except requests.HTTPError as http_err:
>                 # 5xx are usually transient – retry; 4xx are not
>                 if 500 <= resp.status_code < 600 and attempt < max_retries:
>                     wait = 2 ** attempt
>                     log.warning(
>                         "Transient HTTP %s, retry %d/%d after %ds",
>                         resp.status_code, attempt, max_retries, wait,
>                     )
>                     time.sleep(wait)
>                     continue
>                 # re‑raise – caller will see the exception
>                 raise
>             except requests.RequestException as req_err:
>                 # network glitches, timeouts, DNS failures, etc.
>                 if attempt < max_retries:
>                     wait = 2 ** attempt
>                     log.warning("Network error: %s – retry %d/%d after %ds",
>                                 req_err, attempt, max_retries, wait)
>                     time.sleep(wait)
>                     continue
>                 raise
> ```
> 
> Again, I'm not convinced we should do this in the current PR - we could also retry by just calling `bot:status` again. To me, a quick merge is more important, as it is currently very hard to get (complete) status overviews for builds with many targets (i.e. CPU+GPU), which is _exactly_ where this functionality is most essential. It's almost impossible to check if all CPU+GPU combinations for 2025.06 are covered, currently.

Ack.

trz42:
PR has been updated and tested (for tests, see https://github.com/trz42/software-layer/pull/93).
- The update adds a timeout for the API which is configurable. 
- Also uses an access token to benefit from higher REST API limits (5000 vs 60 per hour). REST API rate limit information is obtained and logged.

casparvl (approved):
Looks perfect, tested in https://github.com/casparvl/software-layer/pull/6#issuecomment-3805074258 myself, works like a charm. Great addition @trz42 !

Issue #190, opened 2023-06-23
Add a bot command to immediately kill jobs.


satishskamath:
These are for build jobs that the builder knows are going to fail or wrong.

trz42:
First ideas for UI:
```bash
bot: cancel jobid:25892047
```
Considerations:
- only user who has submitted job should be able to cancel a job
- may need to limit command to specific bot instance

PR #359, opened 2026-01-21
Add new `bot: cancel` command

sondrebr:
See issue #190

The command allows for cancelling multiple jobs in one command, e.g. `bot: cancel jobid:1 jobid:2`. The status table of each job is updated to show that the job was cancelled. A build job can only be cancelled by the job owner (the one who submitted it). This is done by adding the name of the job owner to the job metadata file, and checking that the name matches before cancelling the job.

A new setting is required for the command, 'cancel_command' in the '[buildenv]' section.

One small issue at the moment is that jobs that are known by the job manager get `UNKNOWN` status when they are cancelled, but I guess that can be easily ignored and fixed in another PR if needed.

I have (hopefully) updated the documentation and tests where necessary, although I haven't added any new tests.

First PR, sorry for the single huge commit 😅 Any feedback would be appreciated 🙂


tasks/build.py
```diff
@@ -1358,3 +1366,132 @@ def request_bot_build_issue_comments(repo_name, pr_number):
...
+            continue
+        if any([line.startswith("scancel: error: ") for line in err.split("\n")]):
+            log(f"{fn}(): Unable to cancel job {job_id}.")
+            continue
+
+        log(f"{fn}(): Cancelled job {job_id}")
+
+        # Update job status table
+        dt = datetime.now(timezone.utc)
+        update = f"\n|{dt.strftime("%b %d %X %Z %Y")}|finished|job id `{job_id}` was cancelled|"
```
trz42: Looks like you have to use single quotes for the argument to `strftime`
```suggestion
        update = f"\n|{dt.strftime('%b %d %X %Z %Y')}|finished|job id `{job_id}` was cancelled|"
```

tools/job_metadata.py
```diff
@@ -99,12 +100,14 @@ def create_metadata_file(job, job_id, pr_comment):
...
     pr_comment_id = pr_comment.pr_comment_id
+    job_owner = job.owner
 
     # create _bot_job<jobid>.metadata file in the job's working directory
     bot_jobfile = configparser.ConfigParser()
     bot_jobfile[JOB_PR_SECTION] = {'repo': repo_name,
                                    'pr_number': pr_number,
-                                   'pr_comment_id': pr_comment_id}
+                                   'pr_comment_id': pr_comment_id,
+                                   'job_owner': job_owner}
```
trz42, line 110: I see that we used hardcoded strings as key instead of the constants defined earlier in this file, e.g., `JOB_PR_JOB_OWNER`. We should address this in a follow-up PR.

eessi_bot_event_handler.py
```diff
@@ -507,7 +509,7 @@ def handle_bot_command_help(self, event_info, bot_command):
         help_msg += "\n  - Commands must be sent with a **new** comment (edits of existing comments are ignored)."
         help_msg += "\n  - A comment may contain multiple commands, one per line."
         help_msg += "\n  - Every command begins at the start of a line and has the syntax `bot: COMMAND [ARGUMENTS]*`"
-        help_msg += "\n  - Currently supported COMMANDs are: `help`, `build`, `show_config`, `status`"
+        help_msg += "\n  - Currently supported COMMANDs are: `help`, `build`, `show_config`, `status`, `cancel`"
```
trz42, line 512: Minor suggestion: There are not so many commands, but maybe we could start sorting them alphabetically?

eessi_bot_event_handler.py
```diff
@@ -29,8 +29,8 @@
 
 # Local application imports (anything from EESSI/eessi-bot-software-layer)
 from connections import github
-from tasks.build import check_build_permission, get_node_types, request_bot_build_issue_comments, \
-    submit_build_jobs
+from tasks.build import cancel_jobs, check_build_permission, get_job_ids, get_work_dirs, \
+    get_node_types, request_bot_build_issue_comments, submit_build_jobs
```
trz42: I'd sort imported symbols alphabetically.

tasks/build.py
```diff
@@ -1358,3 +1366,132 @@ def request_bot_build_issue_comments(repo_name, pr_number):
...
+
+    Returns:
+        job_ids (list): valid 'jobid:' arguments
+    """
+    fn = sys._getframe().f_code.co_name
+
+    # Get 'jobid:' arguments
+    job_filter = action_filter.get_filter_by_component(tools_filter.FILTER_COMPONENT_JOB)
+    if not job_filter:
+        log(f"{fn}(): bot: cancel needs at least one 'jobid:' argument.")
```
trz42: Put command in quotes
```suggestion
        log(f"{fn}(): 'bot: cancel' command needs at least one 'jobid:' argument.")
```

tools/filter.py
```diff
@@ -27,7 +27,7 @@
 FILTER_COMPONENT_ARCH = 'architecture'
 FILTER_COMPONENT_EXPORT = 'exportvariable'
 FILTER_COMPONENT_INST = 'instance'
-FILTER_COMPONENT_JOB = 'job'
+FILTER_COMPONENT_JOB = 'jobid'
```
trz42: Should we rename the constant accordingly?
```suggestion
FILTER_COMPONENT_JOBID = 'jobid'
```

tasks/build.py
```diff
@@ -1358,3 +1366,132 @@ def request_bot_build_issue_comments(repo_name, pr_number):
...
+    # Validate job IDs
+    job_ids = []
+    for job_id in job_filter:
+        try:
+            if int(job_id) > 0:
+                job_ids.append(job_id)
+            else:
+                log(f"{fn}(): Invalid job ID: '{job_id}'")
+        except Exception as e:
+            log(f"{fn}(): Invalid job ID: {e}")
```
trz42: Minor suggestion: Don't use single characters as variable names.
```suggestion
        except Exception as err:
            log(f"{fn}(): Invalid job ID: {err}")
```

tasks/build.py
```diff
@@ -1358,3 +1366,132 @@ def request_bot_build_issue_comments(repo_name, pr_number):
...
+        work_dirs (dict): dict mapping each job_id to its work_dir
+    """
+
+    buildenv = get_build_env_cfg(cfg)
+    poll_command = cfg[config.SECTION_JOB_MANAGER][config.JOB_MANAGER_SETTING_POLL_COMMAND]
+    job_name = buildenv[config.BUILDENV_SETTING_JOB_NAME]
+
+    user = os.getenv("USER", None)
+    if user is None:
+        raise Exception("Environment variable $USER is not set.")
```
trz42: We don't need the user name. When polling we can simply use `--me` as argument.

trz42: Since we provide job ids, we might even skip `--me`.

tasks/build.py
```diff
@@ -1358,3 +1366,132 @@ def request_bot_build_issue_comments(repo_name, pr_number):
...
+
+    # squeue only the given job IDs
+    cs_jobs = ",".join(job_ids)
+    command_line = f"{poll_command} --noheader --Format=JobId:0@,WorkDir:0 --user={user} --job={cs_jobs}"
+    if job_name:
+        command_line += f" --name={job_name}"
```
trz42: Don't think we need the name of jobs, because we specify job ids.

eessi_bot_event_handler.py
```diff
@@ -679,6 +681,64 @@ def handle_bot_command_status(self, event_info, bot_command):
...
+            self.log("None of the given jobs are cancellable")
+            return "\n  - No cancellable jobs were given."
+
+        # Log skipped jobs
+        jobs = []
+        for job_id in job_ids:
+            if job_id in work_dirs:
+                jobs.append((job_id, work_dirs.get(job_id)))
+            else:
+                log(f"Skipping job {job_id} - not found")
```
trz42, line 727: Here a new array (`jobs`) is constructed that contains all items in `work_dirs`. Job ids that are not found in `work_dirs` are reported. Minor improvement could be to simply use `work_dirs` instead of constructing a new array? Since we restrict the poll command in `get_work_dirs` to the `job_ids` returned by `get_job_ids`, `work_dirs` can only include these and no others (e.g., new jobs that were submitted between the runs of `get_job_ids` and `get_work_dirs`.

trz42 (changes requested):
Congrats @sondrebr, this looks like a solid PR.

There are a few minor suggestions to consider for this PR. I noted a few other things, which we can address in follow-up PRs.

eessi_bot_event_handler.py
sondrebr: Something for a separate PR, maybe?

eessi_bot_event_handler.py
sondrebr: [Fixed](https://github.com/EESSI/eessi-bot-software-layer/pull/359/commits/d75ec778887d25c6d7f7ebe756e0da7c4bcc1fe1)

tools/filter.py
sondrebr: [Done](https://github.com/EESSI/eessi-bot-software-layer/pull/359/commits/9db5d3280a385b3e2892049e001a7dd62ebc6327)

tasks/build.py
sondrebr: [Fixed](https://github.com/EESSI/eessi-bot-software-layer/pull/359/commits/f452c5d2a6296e0f10098ae1361c8fabe74bddbe)

tasks/build.py
sondrebr: [Fixed](https://github.com/EESSI/eessi-bot-software-layer/pull/359/commits/f452c5d2a6296e0f10098ae1361c8fabe74bddbe)

eessi_bot_event_handler.py
sondrebr: [Done](https://github.com/EESSI/eessi-bot-software-layer/pull/359/commits/30398f165f940a4539d4ffdbb3af5b809ea484ea)

casparvl (commented):
I haven't done a review of the code, but I did test it here https://github.com/casparvl/software-layer/pull/6#issuecomment-3805116129 (cancelled this job here https://github.com/casparvl/software-layer/pull/6#issuecomment-3805118638) and it works perfectly fine! One thing I wasn't able to test is if someone that has permissions to command the bot but was _not_ the submitter of the job can indeed _not_ cancel the job (which was your design goal). But since you need two people for that, maybe it's easier if you test that in some face-to-face session with @trz42 :)

Anyway, nice feature! :)

eessi_bot_event_handler.py
trz42: ok

tools/job_metadata.py
trz42: See #360

eessi_bot_event_handler.py
trz42: See #360

trz42 (commented):
Thanks for the updates @sondrebr ! I'll give it a test on some system. If that succeeds, I'll merge the PR.

Created #360 for remembering the issues to follow up in a subsequent PR.

trz42:
Testing actually revealed a little issue. It tells something about `bot:build` label being set by another user, but this is off. First, no label was set. Second, we used the `bot:label` in the very early days of the bot, but now setting this label doesn't lead to any result (it's just ignored).

See https://github.com/trz42/software-layer/pull/93#issuecomment-3806596674

Can you have a look @sondrebr ?

trz42:
> Testing actually revealed a little issue. It tells something about `bot:build` label being set by another user, but this is off. First, no label was set. Second, we used the `bot:label` in the very early days of the bot, but now setting this label doesn't lead to any result (it's just ignored).
> 
> See [trz42/software-layer#93 (comment)](https://github.com/trz42/software-layer/pull/93#issuecomment-3806596674)
> 
> Can you have a look @sondrebr ?

I think I found a "fix" for this by simply updating a configuration setting. The change below

```diff
diff --git a/app.cfg.example b/app.cfg.example
index a5b1441..f4981d6 100644
--- a/app.cfg.example
+++ b/app.cfg.example
@@ -164,7 +164,7 @@ cancel_command = /usr/bin/scancel
 build_permission = -NOT_ALLOWED_GH_ACCOUNT_NAME-

 # template for comment when user who set a label has no permission to trigger build jobs
-no_build_permission_comment = Label `bot:build` has been set by user `{build_labeler}`, but this person does not have permission to trigger builds
+no_build_permission_comment = GH account `{build_labeler}` is not authorized to trigger or cancel build jobs.

 # whether or not to allow updating the submit options via custom module det_submit_opts
 # Should only be enabled (true) with care because this will result in code from the target
```
leads to a different message as shown in https://github.com/trz42/software-layer/pull/93#issuecomment-3806706904

So my suggestion is to update the setting in `app.cfg.example` and possibly `README.md` in this PR.

We also need to make users aware of this (that they should update the setting or could receive misleading reply) in the notes for the next release (to be created after this PR gets merged).

sondrebr:
Good catch! I have updated the relevant bot comments 👍

trz42 (approved):
Nice work!

PR #370, opened 2026-04-21
[gitlab] Add GitLab connection and EventInfo classes

sondrebr:
First step towards implementing GitLab support for the bot. This PR adds:
- New/updated dependencies (`python-gitlab`, [`PyGHee` with GitLab support](https://github.com/boegel/PyGHee/pull/9))
- New config sections `git` and `gitlab`
- `connections/gitlab.py` for connecting to GitLab
- `tools/git.py` with functions:
  - `get_hosting_platform()` gets the configured Git hosting platform (GitHub/GitLab)
  - `connect_to_host()` performs initial connection to GitHub/GitLab
- `tools/event_info.py` with:
  - `BaseEventInfo` class intended as an "interface" to implement for each Git hosting platform
  - `GitHubEventInfo` class implementing `BaseEventInfo` for GitHub
  - `GitLabEventInfo` class implementing `BaseEventInfo` for GitLab
  - `create_event_info_instance()` creating an instance of the correct EventInfo class
- A `handle_event()` override in the event handler to use the new EventInfo classes

The PR does not any methods for handling GitLab events, but the events will be logged, e.g.:
```
[20260421-T10:53:03] Event received (id: 39c64e78-fdd1-4799-a9c6-568d9b7ffb4b, type: note, action: create), event data logged at [...]
[20260421-T10:53:03] Request verified: signature OK!
[20260421-T10:53:03] WARNING: [event id 39c64e78-fdd1-4799-a9c6-568d9b7ffb4b] No handler found for event type 'note' (action: create) - event was received but left unhandled!
```

`get_hosting_platform()` is intended to be used whenever platform-specific logic is required, to simplify adding support for other platforms in the future (e.g. Forgejo/Codeberg, Gitea, ...).

The `*EventInfo` classes were implemented to handle the differences between the GitHub and GitLab webhooks, such that the necessary information can be retrieved through class properties. For example, `event_info.action` returns the event action. If `GitLabEventInfo` is used, the `action` property converts the action to its GH equivalent before returning, e.g. `update` becomes `edited`.
The classes currently have support for (as far as I can tell) all `event_info` fields used by the bot.
They also have a `__getitem__()` override to support subscripting for compatibility with the existing uses of `event_info` - this version of the bot should therefore still have full GitHub support.

# Usage
Because webhooks signatures are not yet supported by GitLab (though they are [in development](https://gitlab.com/gitlab-org/gitlab/-/work_items/19367)), it is necessary to run a custom Smee server to sign the webhooks. I have a fork [here](https://github.com/sondrebr/smee.io) which should sign the events in the same way as they will be signed on GitLab in the future. I have also added a "Quickstart" section on top of the README in the fork with instructions on how to run the server.

To "use" the bot in its current state on GitLab, you need to do the following:
- Install the requirements (`pip install -r requirements.txt`)
Ensure the correct version of PyGHee (commit `c5e1063...`) is installed.
- Create a webhook on a GitLab project (`Settings` -> `Webhooks` -> `Add new`)
The webhook should have the `Comments` and `Merge request events` triggers, and must be configured with a secret token. The secret token can be generated e.g. by running `python -c "import secrets; print(secrets.token_hex())"`, and must be stored as the environment variable `GITLAB_WEBHOOK_SECRET_TOKEN`. The webhooks must also be set up to send events to the custom Smee server.
- Create a project access token for the same GitLab project (`Settings` -> `Access tokens` -> `Add new token`)
In the future, the `api` scope and `Planner` will be required for the bot to be able to write comments. For this PR, however, it just needs to be able to connect to GitLab, so the `read_api` scope and `Guest` role should be sufficient. The access token must be stored as the environment variable `GITLAB_PROJECT_ACCESS_TOKEN`.
- Configure the `git` and `gitlab` sections in `app.cfg`
In the `git` section, `hosting_platform = gitlab` must be set (if it is set to `github` instead, the bot should function normally).
In the `gitlab` section, the `api_timeout`, `bot_name`, and `instance_url` options must be set. For the `instance_url` option, you must enter the base URL of your GitLab instance, which is used along with the access token to connect to GitLab.

### DIFF
--- a/connections/gitlab.py
+++ b/connections/gitlab.py
@@ -1,0 +1,87 @@
+# This file is part of the EESSI build-and-deploy bot,
+# see https://github.com/EESSI/eessi-bot-software-layer
+#
+# The bot helps with requests to add software installations to the
+# EESSI software layer, see https://github.com/EESSI/software-layer
+#
+# author: Sondre Bergsvaag Risanger (@sondrebr)
+#
+# license: GPLv2
+#
+
+# Standard library imports
+import os
+
+# Third party imports (anything installed into the local Python environment)
+import gitlab
+
+# Local application imports (anything from EESSI/eessi-bot-software-layer)
+from tools import config, logging
+
+
+_gl = None
+
+
+def verify_connection(gl):
+    """
+    Verifies connection to GitLab. Exits if verification fails.
+
+    Args:
+        Instance of Gitlab
+
+    Returns:
+        None (implicit)
+    """
+    try:
+        # auth tests the instance's credentials by retrieving the access token user
+        gl.auth()
+        if type(gl.user) is not gl._objects.CurrentUser:
+            raise Exception("'user' attribute of Gitlab instance is not of type 'CurrentUser'.")
+    except Exception as err:
+        logging.error(f"Failed to verify GitLab connection: {err}")
+
+
+def connect():
+    """
+    Creates a Gitlab instance, then verifies the connection.
+
+    Args:
+        No arguments
+
+    Returns:
+        None (implicit)
+    """
+    global _gl
+    cfg = config.read_config()
+    gitlab_cfg = cfg[config.SECTION_GITLAB]
+    timeout = int(gitlab_cfg.get(config.GITLAB_SETTING_API_TIMEOUT, 10))
+    url = gitlab_cfg.get(config.GITLAB_SETTING_INSTANCE_URL)
+
+    access_token = os.getenv('GITLAB_PROJECT_ACCESS_TOKEN')
+    if access_token is None:
+        logging.error("GitLab token is not available via $GITLAB_PROJECT_ACCESS_TOKEN!")
+    else:
+        del os.environ['GITLAB_PROJECT_ACCESS_TOKEN']
+
+    _gl = gitlab.Gitlab(url, access_token)
+    _gl.timeout = timeout
+    _gl.retry_transient_errors = True
+    verify_connection(_gl)
+
+
+def get_instance():
+    """
+    Returns a Gitlab instance. Creates an instance if one does not exist,
+    otherwise verifies the existing instance.
+
+    Args:
+        No arguments
+
+    Returns:
+        Instance of Gitlab
+    """
+    if not _gl:
+        connect()
+    else:
+        verify_connection(_gl)
+    return _gl
--- a/connections/gitlab.py
+++ b/connections/gitlab.py
@@ -27,7 +27,7 @@ def verify_connection(gl):
     Verifies connection to GitLab. Exits if verification fails.
 
     Args:
-        Instance of Gitlab
+        Instance of gitlab.Gitlab (from python-gitlab)
 
     Returns:
         None (implicit)
@@ -36,14 +36,14 @@ def verify_connection(gl):
         # auth tests the instance's credentials by retrieving the access token user
         gl.auth()
         if type(gl.user) is not gl._objects.CurrentUser:
-            raise Exception("'user' attribute of Gitlab instance is not of type 'CurrentUser'.")
+            raise Exception("'user' attribute of Gitlab class instance is not of type 'CurrentUser'.")
     except Exception as err:
         logging.error(f"Failed to verify GitLab connection: {err}")
 
 
 def connect():
     """
-    Creates a Gitlab instance, then verifies the connection.
+    Creates a gitlab.Gitlab instance (from python-gitlab), then verifies the connection to GitLab.
 
     Args:
         No arguments
@@ -71,14 +71,14 @@ def connect():
 
 def get_instance():
     """
-    Returns a Gitlab instance. Creates an instance if one does not exist,
+    Returns a gitlab.Gitlab instance. Creates an instance if one does not exist,
     otherwise verifies the existing instance.
 
     Args:
         No arguments
 
     Returns:
-        Instance of Gitlab
+        Instance of gitlab.Gitlab (from python-gitlab)
     """
     if not _gl:
         connect()
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -38,7 +38,7 @@ from tools.args import event_handler_parse
 from tools.commands import EESSIBotCommand, EESSIBotCommandError, \
     contains_any_bot_command, get_bot_command
 from tools.event_info import create_event_info_instance
-from tools.git import connect_to_host, get_hosting_platform
+from tools.git import connect_to_git_hosting_platform, get_git_hosting_platform
 from tools.permissions import check_command_permission
 from tools.pr_comments import ChatLevels, create_comment
 
@@ -141,7 +141,7 @@ class EESSIBotSoftwareLayer(PyGHee):
         EESSIBotSoftwareLayer constructor. Calls constructor of PyGHee and
         initializes some configuration settings.
         """
-        event_source = get_hosting_platform()
+        event_source = get_git_hosting_platform()
         super(EESSIBotSoftwareLayer, self).__init__(event_source, *args, **kwargs)
 
         self.cfg = config.read_config()
@@ -858,8 +858,8 @@ def main():
         print("Configuration check: FAILED")
         sys.exit(1)
 
-    # Connect to Git hosting platform
-    connect_to_host()
+    # Verify that the event handler is able to connect to the Git hosting platform
+    connect_to_git_hosting_platform()
 
     if opts.file:
         app = create_app(klass=EESSIBotSoftwareLayer)
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -37,6 +37,7 @@ from tools import config
 from tools.args import event_handler_parse
 from tools.commands import EESSIBotCommand, EESSIBotCommandError, \
     contains_any_bot_command, get_bot_command
+from tools.event_info import create_event_info_instance
 from tools.git import connect_to_host, get_hosting_platform
 from tools.permissions import check_command_permission
 from tools.pr_comments import ChatLevels, create_comment
@@ -164,6 +165,21 @@ class EESSIBotSoftwareLayer(PyGHee):
             msg = msg % args
         msg = "[%s]: %s" % (funcname, msg)
         log(msg, log_file=self.logfile)
+
+    def handle_event(self, event_info, log_file=None):
+        """
+        Create EventInfo instance using event_info,
+        then pass that to PyGHee's handle_event method.
+
+        Args:
+            event_info (dict): event received by event_handler
+            log_file (string): path to log messages to
+
+        Returns:
+            None (implicit)
+        """
+        event_info_object = create_event_info_instance(event_info)
+        super().handle_event(event_info_object, log_file)
 
     def handle_issue_comment_event(self, event_info, log_file=None):
         """
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -37,6 +37,7 @@ from tools import config
 from tools.args import event_handler_parse
 from tools.commands import EESSIBotCommand, EESSIBotCommandError, \
     contains_any_bot_command, get_bot_command
+from tools.git import connect_to_host, get_hosting_platform
 from tools.permissions import check_command_permission
 from tools.pr_comments import ChatLevels, create_comment
 
@@ -95,12 +96,18 @@ REQUIRED_CONFIG = {
         config.DOWNLOAD_PR_COMMENTS_SETTING_PR_DIFF_TIP],          # required
     config.SECTION_EVENT_HANDLER: [
         config.EVENT_HANDLER_SETTING_LOG_PATH],                    # required
+    config.SECTION_GIT: [
+        config.GIT_SETTING_HOSTING_PLATFORM],                      # required
     config.SECTION_GITHUB: [
-        config.GITHUB_SETTING_API_TIMEOUT,                         # required
-        config.GITHUB_SETTING_APP_ID,                              # required
-        config.GITHUB_SETTING_APP_NAME,                            # required
-        config.GITHUB_SETTING_INSTALLATION_ID,                     # required
-        config.GITHUB_SETTING_PRIVATE_KEY],                        # required
+        config.GITHUB_SETTING_API_TIMEOUT,                         # required for github
+        config.GITHUB_SETTING_APP_ID,                              # required for github
+        config.GITHUB_SETTING_APP_NAME,                            # required for github
+        config.GITHUB_SETTING_INSTALLATION_ID,                     # required for github
+        config.GITHUB_SETTING_PRIVATE_KEY],                        # required for github
+    config.SECTION_GITLAB: [
+        config.GITLAB_SETTING_API_TIMEOUT,                         # required for gitlab
+        config.GITLAB_SETTING_BOT_NAME,                            # required for gitlab
+        config.GITLAB_SETTING_INSTANCE_URL],                       # required for gitlab
     # the poll interval setting is required for the alternative job handover
     # protocol (delayed_begin)
     config.SECTION_JOB_MANAGER: [
@@ -133,7 +140,8 @@ class EESSIBotSoftwareLayer(PyGHee):
         EESSIBotSoftwareLayer constructor. Calls constructor of PyGHee and
         initializes some configuration settings.
         """
-        super(EESSIBotSoftwareLayer, self).__init__(*args, **kwargs)
+        event_source = get_hosting_platform()
+        super(EESSIBotSoftwareLayer, self).__init__(event_source, *args, **kwargs)
 
         self.cfg = config.read_config()
         event_handler_cfg = self.cfg[config.SECTION_EVENT_HANDLER]
@@ -833,7 +841,9 @@ def main():
     else:
         print("Configuration check: FAILED")
         sys.exit(1)
-    github.connect()
+
+    # Connect to Git hosting platform
+    connect_to_host()
 
     if opts.file:
         app = create_app(klass=EESSIBotSoftwareLayer)
--- a/eessi_bot_event_handler.py
+++ b/eessi_bot_event_handler.py
@@ -168,6 +168,7 @@ class EESSIBotSoftwareLayer(PyGHee):
 
     def handle_event(self, event_info, log_file=None):
         """
+        Override of PyGHee's handle_event method.
         Create EventInfo instance using event_info,
         then pass that to PyGHee's handle_event method.
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,5 +15,5 @@ PyGithub
 python-gitlab
 Waitress>=3.0.1 # required to fix vulnerabilities detected by scorecards
 cryptography>=44.0.1 # required to fix vulnerabilities detected by scorecards
-PyGHee @ git+https://github.com/boegel/PyGHee.git@bd4ab122206c446bd4fe554e1344d26a64b232d3 # Pin commit adding GL support
+PyGHee @ git+https://github.com/boegel/PyGHee.git@c5e10632a45db5ca94f5cbf87ac7a90a2064e8fd # Pin commit with GL support
 retry
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,11 +7,13 @@
 # author: Bob Droege (@bedroge)
 # author: Kenneth Hoste (@boegel)
 # author: Thomas Roeblitz (@trz42)
+# author: Sondre Bergsvaag Risanger (@sondrebr)
 #
 # license: GPLv2
 #
 PyGithub
+python-gitlab
 Waitress>=3.0.1 # required to fix vulnerabilities detected by scorecards
 cryptography>=44.0.1 # required to fix vulnerabilities detected by scorecards
-PyGHee>=0.0.3
+PyGHee @ git+https://github.com/boegel/PyGHee.git@bd4ab122206c446bd4fe554e1344d26a64b232d3 # Pin commit adding GL support
 retry
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,8 @@
 # license: GPLv2
 #
 PyGithub
-python-gitlab
+python-gitlab==6.5.0;python_version=="3.9" # Last version with Python 3.9 support
+python-gitlab==8.3.0;python_version>="3.10" # Most recent version on 2026-05-04
 Waitress>=3.0.1 # required to fix vulnerabilities detected by scorecards
 cryptography>=44.0.1 # required to fix vulnerabilities detected by scorecards
 PyGHee @ git+https://github.com/boegel/PyGHee.git@c5e10632a45db5ca94f5cbf87ac7a90a2064e8fd # Pin commit with GL support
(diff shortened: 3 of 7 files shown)


sondrebr:
Had to temporarily switch the base branch to get rid of some old commits 🤦 Ready for review now.

requirements.txt
```diff
@@ -7,11 +7,13 @@
 # author: Bob Droege (@bedroge)
 # author: Kenneth Hoste (@boegel)
 # author: Thomas Roeblitz (@trz42)
+# author: Sondre Bergsvaag Risanger (@sondrebr)
 #
 # license: GPLv2
 #
 PyGithub
+python-gitlab
```
trz42: Can this be pinned to a specific version too?

connections/gitlab.py
```diff
@@ -0,0 +1,87 @@
...
+
+_gl = None
+
+
+def verify_connection(gl):
+    """
+    Verifies connection to GitLab. Exits if verification fails.
+
+    Args:
+        Instance of Gitlab
```
trz42: Tiny comment. Use consistent spelling of GitLab.

connections/gitlab.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Returns:
+        None (implicit)
+    """
+    try:
+        # auth tests the instance's credentials by retrieving the access token user
+        gl.auth()
+        if type(gl.user) is not gl._objects.CurrentUser:
+            raise Exception("'user' attribute of Gitlab instance is not of type 'CurrentUser'.")
```
trz42: Tiny comment. Use consistent spelling of GitLab.

connections/gitlab.py
```diff
@@ -0,0 +1,87 @@
...
+    except Exception as err:
+        logging.error(f"Failed to verify GitLab connection: {err}")
+
+
+def connect():
+    """
+    Creates a Gitlab instance, then verifies the connection.
```
trz42: Tiny comment. Use consistent spelling of GitLab.

connections/gitlab.py
```diff
@@ -0,0 +1,87 @@
...
+    cfg = config.read_config()
+    gitlab_cfg = cfg[config.SECTION_GITLAB]
+    timeout = int(gitlab_cfg.get(config.GITLAB_SETTING_API_TIMEOUT, 10))
+    url = gitlab_cfg.get(config.GITLAB_SETTING_INSTANCE_URL)
+
+    access_token = os.getenv('GITLAB_PROJECT_ACCESS_TOKEN')
+    if access_token is None:
+        logging.error("GitLab token is not available via $GITLAB_PROJECT_ACCESS_TOKEN!")
+    else:
+        del os.environ['GITLAB_PROJECT_ACCESS_TOKEN']
```
trz42, line 64: Nice!

connections/gitlab.py
```diff
@@ -0,0 +1,87 @@
...
+
+    _gl = gitlab.Gitlab(url, access_token)
+    _gl.timeout = timeout
+    _gl.retry_transient_errors = True
+    verify_connection(_gl)
+
+
+def get_instance():
+    """
+    Returns a Gitlab instance. Creates an instance if one does not exist,
```
trz42: Tiny comment. Use consistent spelling of GitLab.

tools/config.py
```diff
@@ -207,9 +216,12 @@ def check_cfg_settings(req_settings, path="app.cfg"):
     """
     # TODO argument path is not being used
     cfg = read_config()
+    git_host = get_hosting_platform(cfg)
     # iterate over keys in req_settings which correspond to sections ([name])
     # in the configuration file (.ini format)
     for section in req_settings.keys():
+        if git_host and (section in SUPPORTED_HOSTS) and (section != git_host):
```
trz42: Minor comment. Maybe `SUPPORTED_HOSTS` could be renamed to something like `SUPPORTED_GIT_HOSTS`.

tools/config.py
```diff
@@ -207,9 +216,12 @@ def check_cfg_settings(req_settings, path="app.cfg"):
...
+            continue
```
trz42, line 225: Maybe add a comment why the section is skipped.

eessi_bot_event_handler.py
```diff
@@ -833,7 +857,9 @@ def main():
     else:
         print("Configuration check: FAILED")
         sys.exit(1)
-    github.connect()
+
+    # Connect to Git hosting platform
+    connect_to_host()
```
trz42: Maybe name could tell a little more what it is supposed to do, for example,

`connect_to_git_hosting_platform`

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+    retrieve it instead of subscripting/using the event_info dict.
+    """
+    def __init__(self, event_info):
+        if self.__class__ is BaseEventInfo:
+            err_msg = "Do not use this base class directly. "
+            err_msg += "Please use one of its subclasses instead."
+            raise NotImplementedError(err_msg)
+        self.event_info = event_info
+
+    # Do not override - implements subscripting for compatibility
```
trz42, line 43: We could enforce the "Do not override" via something like

```python
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        if "__getitem__" in cls.__dict__:
            raise TypeError(f"{cls.__name__} must not override __getitem__")
```

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+    def pr_merged_status(self):
+        raise NotImplementedError()
+
+    @cached_property
+    def pr_url(self):
+        raise NotImplementedError()
+
+    @cached_property
+    def repo_name(self):
+        raise NotImplementedError()
```
trz42, line 101: Wonder if we could do this a little more "elegant", for example:

```python
def _not_implemented(name):
    @cached_property
    def prop(self):
        raise NotImplementedError(f"{name} is not implemented for {type(self).__name__}")
    prop.__set_name__ = lambda self, owner, _: None  # keeps cached_property happy
    return prop

_ABSTRACT_PROPERTIES = [
    "action", "comment_body", "comment_created_by", "comment_body",
    "comment_updated_by", "discussion_id", "event_id", "event_type",
    "label_name", "pr_number", "pr_merged_status", "pr_url", "repo_name",
]

class BaseEventInfo():
    def __init__(self, event_info):
        if self.__class__ is BaseEventInfo:
            err_msg = "Do not use this base class directly. "
            err_msg += "Please use one of its subclasses instead."
            raise NotImplementedError(err_msg)
        self.event_info = event_info

    def __getitem__(self, key):
        return self.event_info[key]

for _name in _ABSTRACT_PROPERTIES:
    setattr(BaseEventInfo, _name, _not_implemented(_name))
```

eessi_bot_event_handler.py
```diff
@@ -157,6 +166,21 @@ def log(self, msg, *args):
         msg = "[%s]: %s" % (funcname, msg)
         log(msg, log_file=self.logfile)
 
+    def handle_event(self, event_info, log_file=None):
+        """
+        Create EventInfo instance using event_info,
+        then pass that to PyGHee's handle_event method.
```
trz42, line 173: So this overrides PyGHee's `handle_event`, right?

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+
+    Returns:
+        Instance of BaseEventInfo subclass
+    """
+    git_host = get_hosting_platform()
+    if git_host == GITHUB:
+        new_event_info = GitHubEventInfo(event_info)
+    elif git_host == GITLAB:
+        new_event_info = GitLabEventInfo(event_info)
+    return new_event_info
```
trz42: If `git_host` is neither GITHUB nor GITLAB, the `return` will raise an `UnboundLocalError`. How can we avoid that, or differently asked how could we ensure if `get_hosting_platform` changes that we change the code here too?

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+        created_by_id = self._object_attributes["author_id"]
+        triggered_by_id = self._request_body["user"]["id"]
+        # GL events only include the username of the user who triggered the event
+        if triggered_by_id == created_by_id:
+            created_by = self._request_body["user"]["username"]
+        else:
+            gl = gitlab.get_instance()
+            user = gl.users.get(created_by_id)
+            created_by = user.username
+        return created_by
```
trz42, line 230: Don't fully understand why we're not interested in who has triggered the event as in who has created or updated a comment in a merge request?

Looking at https://docs.gitlab.com/user/project/integrations/webhook_events/#comment-on-a-merge-request it seems to me that `raw_request_body["user"]["username"]` is what we want to use?

`created_by_id` - corresponds to the user who created the comment?
`triggered_by_id` - corresponds to the user who created the comment?

Aren't they always the same?

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @cached_property
+    def comment_updated_by(self):
+        return self._request_body["user"]["username"]
```
trz42: Do we actually need to know who has updated a comment?

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+    def label_name(self):
+        return self._request_body["label"]["name"]
+
+    @cached_property
+    def pr_number(self):
+        if self.event_type == "pull_request":
+            pr_num = self._request_body["pull_request"]["number"]
+        else:
+            pr_num = self._request_body["issue"]["number"]
+        return pr_num
```
trz42: Should we have two methods for each base type - PR and ISSUE?

Now, we would use `obj.pr_number()` somewhere in the code, but depending the type of `obj` we either get a PR number or an ISSUE number.

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+        return self._request_body["comment"]["user"]["login"]
+
+    @cached_property
+    def comment_updated_by(self):
+        return self._request_body["sender"]["login"]
+
+    @cached_property
+    def discussion_id(self):
+        # Not applicable for GitHub
+        return None
```
trz42: Wonder if we should have a public methods in the interface that is only relevant for a specific hosting platform.

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+    def pr_merged_status(self):
+        return self._request_body["pull_request"]["merged"]
+
+    @cached_property
+    def pr_url(self):
+        if self.event_type == "pull_request":
+            url = self._request_body["pull_request"]["html_url"]
+        else:
+            url = self._request_body["issue"]["html_url"]
+        return url
```
trz42: Similarly to the comment about `pr_number`, we might want two methods here too.

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+    @cached_property
+    def event_type(self):
+        gl_event_type = self.event_info["type"]
+        return self._EVENT_TYPE_MAP.get(gl_event_type, self._UNKNOWN)
+
+    @cached_property
+    def label_name(self):
+        # GH sends one event per label while GL sends one event with all label changes
+        # Set manually and call handle_pull_request_labeled_event once per label?
+        raise NotImplementedError()
```
trz42: So far, we only use a single label `bot:deploy` and thus it may not make much difference for GL and GH. Maybe we should refactor the UI to allow for another command such as
```
bot: deploy [args_similar_to_bot_build_cmd]
```
We would avoid complications due to different ways how events are created for changes in labels.

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @cached_property
+    def pr_number(self):
+        if self.event_type == "pull_request":
+            pr_iid = self._object_attributes["iid"]
+        else:
+            pr_iid = self._request_body["merge_request"]["iid"]
+        return pr_iid
```
trz42, line 298: How could it happen that `self.event_type` equals `pull_request` and thus we have to obtain the `pr_number` differently than when it is not a pull request?

This is a bit confusing. Maybe a comment which explains the reason for the differences is all what is needed.

tools/event_info.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @cached_property
+    def pr_merged_status(self):
+        if self.event_type == "pull_request":
+            state = self._object_attributes["state"]
+        else:
+            state = self._request_body["merge_request"]["state"]
+        return state == "merged"
```
trz42, line 306: Different processing depending on `event_type` should be explained.

tools/event_info.py
trz42: `iid` seems to be the internal id of the merge request. Is this really what we want here?

See https://docs.gitlab.com/user/project/integrations/webhook_events/#merge-request-events

Hmm, looking at the example in https://docs.gitlab.com/user/project/integrations/webhook_events/#complete-payload-example `iid` could be what we need.

trz42 (changes requested):
This is a solid PR tackling the first step towards supporting repositories hosted on GitLab and GitHub.

What's a bit missing is explaining some kind of data model that is used to form the interface defined by BaseEventInfo. This doesn't need to cover all kinds of events and actions GitHub and GitLab support, but only the ones we currently use in the bot which only supports GitHub.

While the difference in how events are generated for labels might trigger a more substantial UI change (replacing setting the label `bot:deploy` with a new command `bot:deploy [ARGS]`), we might live with a workaround as the bot only supports one specific label (`bot:deploy`).

More comments (mostly suggesting changes ... not really requesting them) are provided with the code.

requirements.txt
trz42: Can one pin it to different versions for different python versions?

tools/event_info.py
trz42: For now it would be sufficient if the label `bot:deploy` was set and if so return it or return `None`.

tools/event_info.py
trz42:
```suggestion
    git_host = get_hosting_platform()
    if git_host == GITHUB:
        return GitHubEventInfo(event_info)
    elif git_host == GITLAB:
        return GitLabEventInfo(event_info)
    return None
```

requirements.txt
sondrebr: Done: ab02d4dd35190b742213a0af0612fcd33f4b3dc8
Python 3.9 and 3.10 onwards have separate pins because python-gitlab dropped support for Python 3.9 with version 7.0.0.

connections/gitlab.py
sondrebr: `Gitlab` refers to the GitLab client class of python-gitlab, similar to how `Github` refers to the GitHub client class of PyGithub in `connections/github.py`.

I guess it would make sense to specify that it is a class, e.g. `Instance of Gitlab class`, to avoid confusion with actual GitLab instances?

tools/event_info.py
sondrebr: The intention is to not have to check which platform the bot is configured for before accessing EventInfo properties, but I agree it might seem weird. Seeing as no part of the bot currently uses the discussion ID, I think I'll remove the property from this PR and reintroduce it later if/when it makes sense.

tools/event_info.py
sondrebr: I ended up putting them in the same `pr_number` property because the bot always sets the issue number to a variable named `pr_number` (see for example `pr_number = request_body['issue']['number']` in `EESSIBotSoftwareLayer.handle_issue_comment_event()`), but looking at it now, it makes more sense to separate it into two properties.

tools/event_info.py
sondrebr: Thinking about it now, the property name should be renamed - `event_triggered_by` more accurately describes what this field returns.

tools/event_info.py
sondrebr: GL for some reason puts the event field under `object_attributes`. In this case, events from comments on merge requests will have information about the MR in a `merge_request` field, while merge request events will have the same information in `object_attributes`...

As for `iid`, that is the internal (project-wide) ID, which is what we want. The `id` field contains what I assume is the instance-wide ID, which AFAIK is not needed for the bot.

tools/event_info.py
sondrebr: Added a comment in 611022a7173723a8092949846b2b3c8995084cd3 addressing `object_attributes` and how it makes the processing change depending on the event type.

tools/event_info.py
sondrebr: Added a comment in 611022a7173723a8092949846b2b3c8995084cd3 addressing `object_attributes` and how it makes the processing change depending on the event type.

eessi_bot_event_handler.py
sondrebr: Correct. Updated the docstring in 611022a7173723a8092949846b2b3c8995084cd3 to specify that it overrides the method.

tools/config.py
sondrebr: Added a comment in 611022a7173723a8092949846b2b3c8995084cd3.

tools/config.py
sondrebr: Done: 9e7f6eb27426c704d3cbf93858c6283ed6ca661f

eessi_bot_event_handler.py
sondrebr: Agreed. I changed the name in 9e7f6eb27426c704d3cbf93858c6283ed6ca661f, and also changed `get_hosting_platform` to `get_git_hosting_platform` to match.

connections/gitlab.py
sondrebr: > I guess it would make sense to specify that it is a class, e.g. `Instance of Gitlab class`, to avoid confusion with actual GitLab instances?

Addressed this in 4155c30e9d16d725e2b7a99b67f4cc5d55801ce9.

tools/event_info.py
sondrebr: Done: 5e5259fad22514527f044ae108a112598e6fe5f1. Didn't know about `__init_subclass__`, pretty handy 👍

tools/event_info.py
sondrebr: Renamed it in 6996b9ea4bd605154c6d443f142436b96d019351

tools/event_info.py
sondrebr: 91c4860e1c570b96cd26feb48e384e723ac06ed7 removes the property.

tools/event_info.py
sondrebr: Done: f8391b0f0abcd56b424318915e1749fd73e64f10. Hopefully the comments explain it well enough 😅

tools/event_info.py
sondrebr: Done: 8502631c5001cb9ed76fd4e044386407b16cbe14. The GL implementation might look a bit weird because they use the `merge_request` or `issue` field depending on the comment type (and the comment trigger also sends events for commit comments), while GH uses the `issue` field for both issue and PR comments 😶

tools/event_info.py
sondrebr: Done: 8502631c5001cb9ed76fd4e044386407b16cbe14. Also check my reply on the other comment (I tried adding a link here but it doesn't seem to actually go to the reply...).

tools/event_info.py
sondrebr: I expanded the comment to hopefully address this: 611022a7173723a8092949846b2b3c8995084cd3. It is also worth noting that even for comment update events (which are skipped), the username of the original author is logged (early in `EESSIBotSoftwareLayer.handle_issue_comment_event`, before the `action == 'created'` check).

tools/event_info.py
sondrebr: I want to keep it as-is, at least for now. The current version is IMO more readable, and makes it easier to add docstrings to the properties later.

tools/event_info.py
sondrebr: Good point about changes in `get_git_hosting_platform`. I used your suggestion in 373e6db708b31f85216faca6f392672e20c6abc9.

sondrebr:
> What's a bit missing is explaining some kind of data model that is used to form the interface defined by BaseEventInfo. This doesn't need to cover all kinds of events and actions GitHub and GitLab support, but only the ones we currently use in the bot which only supports GitHub.

I have these notes (which I've cleaned up a bit) from when I prepared to start working on GL support for the bot. Hopefully they show the basis for `BaseEventInfo`, which I think covers all fields below. Note that with the `pr_url` property now split up into `issue_url` and `pr_url`, `issue_url` will be used, while `pr_url` as far as I can tell will be unused. Maybe a use case pops up, or we can remove it?

For the section documenting `raw_request_body`, the entries have the variable name used in the bot on the left-hand side, showing (to some degree) how the field is used. The `_.{field}` notation shows which field it is in the body, e.g. `_.comment.body` corresponds to `event_info['raw_request_body']['comment']['body']`.

---------

# Where fields are used
## `raw_request_body`
- `update_pr_comment()` function in `tools/pr_comments.py` (unused?)
	- `comment_new = _.comment.body`
	- `repo_name = _.repository.full_name`
	- `pr_number = _.issue.number`
	- `issue_id = _.comment.id`
- `deploy_built_artefacts()` function of `tasks/deploy.py`
	- `labeler = _.sender.login`
	- `repo_name = _.repository.full_name`
- `prepare_jobs()` function of `tasks/build.py`
	- `job_owner = _.sender.login`
- `check_build_permission()` function of `tasks/build.py`
	- `build_labeler = _.sender.login`
	- `repo_name = _.repository.full_name`
- Different `handle_` methods in `EESSIBotSoftwareLayer` class
	- `issue_url = _.issue.url`
	- `action = _.action`
	- `sender = _.sender.login`
	- `owner = _.comment.user.login`
	- `repo_name = _.repository.full_name`
	- `pr_number = _.issue.number`
	- `comment_received = _.comment.body`
	- `label = _.label.name`
		- GL has no "labeled" action - can be identified by checking for `_.changes.labels`
	- `pr_number = _.pull_request.number`
	- `mergedOrClosed = _.pull_request.merged`
	
### Summary - Unique used fields in `raw_request_body` -> property name in `EventInfo`
- Misc. fields
  - `_.action` -> `action`
  - `_.repository.full_name` -> `repo_name`
  - `_.sender.login` -> `event_triggered_by`
- PR/issue-related fields
  - `_.pull_request.number` -> `pr_number`
  - `_.pull_request.merged` -> `pr_merged_status`
  - `_.issue.number` -> `issue_number`
  - `_.issue.url` -> `issue_url`
  - `_.label.name` -> `label_name`
- Comment-related fields
  - `_.comment.body` -> `comment_body`
  - `_.comment.id` -> `comment_id`
  - `_.comment.user.login` -> `comment_created_by`

## `action` (from body) -> `action` in `EventInfo`
- `handle_issue_comment_event()` method of `EESSIBotSoftwareLayer` class
	- Comment is processed only if action is `created`
- `handle_installation_event()` method of `EESSIBotSoftwareLayer` class
	- Logs user & action
- `handle_pull_request_event()` method of `EESSIBotSoftwareLayer` class
	- Used to get `handle_pull_request_{action}_event()` method
		- Actions with methods:
			- `labeled`
			- `opened`
			- `closed`
## `id` (from header) -> `event_id` in `EventInfo`
- `create_pr_dir()` function of `tasks/build.py`
	- Used to create job directory (`**/event_{id}/`)

## `type` (from header) -> `event_type` in `EventInfo`
- `handle_event()` method of `PyGHee` class
	- Used to get `handle_{type}_event()` method
		- Installation (N/A for GL)
		- Issue comment
		- Pull request

tools/event_info.py
trz42: ok

trz42 (approved):
Looks good. Thanks a lot for this work @sondrebr 

Will open an issue or two so we don't forget about updating documentation (README.md) and add/change unit tests.

PR #376, opened 2026-05-28
[gitlab] Add PRComment classes and GitLab support for `bot: help` command

sondrebr:
Another step in implementing GitLab support for the bot, most notably making it support the `bot: help` command. Changes include:
- In `tools/pr_comments.py`:
  - Existing `PRComment` class renamed to `PRCommentInfo`
  - Added `BasePRComment` class, "interface" to implement for each Git hosting platform to support creating/getting/editing PR comments
  - Added `GitHubPRComment` class, GitHub implementation of `BasePRComment`
  - Added `GitLabPRComment` class, GitLab implementation of `BasePRComment`
  - Added `create_pr_comment_instance()` function creating an instance of the correct `PRComment` class
  - Changed the `create_comment()` function to use the `PRComment` classes
- In `tools/commands.py`:
  - Added `ALL_COMMANDS` constant (list containing all existing bot commands)
  - Added `SUPPORTED_COMMANDS_PER_GIT_HOST` constant (dictionary listing supported/implemented commands per Git hosting platform)
  - Added `get_supported_commands()` function retrieving the supported commands for the configured Git hosting platform
- In `eessi_bot_event_handler.py`:
  - Change `handle_issue_comment_event()` method to use the EventInfo properties and `get_app_name()`
  - Add `handle_note_event = handle_issue_comment_event` to make the bot process GitLab comments
  - Change `handle_bot_command()` method to check if command is supported on the configured Git hosting platform
  - Change `bot: help` response message to include a list of unsupported commands (if any)
- Added `get_app_name()` to `tools/git.py` to get the app/bot name (separate settings for GitHub/GitLab)
- Updated PyGHee dependecy
- Typo fix in EventInfo (I had used `notable_type`  instead of `noteable_type` 🤦)
- EventInfo and PRComment types (referring to subclasses of BaseEventInfo and BasePRComment, respectively)

# Usage
To use the bot in its current state on GitLab, follow the Usage section in the description of #370, except:
- When updating PyGHee, check that commit `514bdf6...` is installed.
- The GitLab access token must have the `api` scope and the *Reporter* role (not Planner!) for the bot to be able to post comments.
- Setting up a custom Smee server is no longer required *if* you use GitLab.com or GitLab CE/EE 19.0 or above, as they support signing tokens on webhooks. In that case, GitLab will generate a signing token for you, which must be stored as the `GITLAB_WEBHOOK_SECRET_TOKEN` environment variable. Make sure to include the `whsec_` prefix.
  - Side note if using GitLab.com: AFAIK it is not possible to create project access tokens on the free plan. One can however create a service account in the group/project and generate an access token for the service account.
- If using a GitLab instance without support for signing tokens, you still need to use a custom Smee server. Additionally, you need to use a webhook secret similar to the signing tokens GitLab generates, consisting of 32 bytes encoded in base64, prefixed with `whsec_`. These can be generated e.g. by running `python -c "import base64, secrets; print('whsec_' + base64.b64encode(secrets.token_bytes(32)).decode())"`.

### DIFF
--- a/tools/pr_comments.py
+++ b/tools/pr_comments.py
@@ -295,16 +295,14 @@ class GitHubPRComment(BasePRComment):
         return None
 
     def get(self):
-        if not self.id:
-            raise Exception("'id' must be set to get a comment.")
+        self._require_id()
         self._comment_obj = retry_call(self._pr_obj.get_issue_comment, fargs=[self.id],
                                        exceptions=Exception, tries=5, delay=1, backoff=2, max_delay=30)
         if self._comment_obj:
             self.body = self._comment_obj.body
 
     def create(self):
-        if not self.body:
-            raise Exception("'body' must be set to create a comment.")
+        self._require_body()
         if self.id:
             # Safeguard: id should not be set when body is initialized, but guard
             # against accidental duplicate comments if id was set after initialization.
@@ -315,8 +313,7 @@ class GitHubPRComment(BasePRComment):
             self.id = self._comment_obj.id
 
     def edit(self, new_body):
-        if not self.id:
-            raise Exception("'id' must be set to edit a comment.")
+        self._require_id()
         # Ensure comment object is present
         if not self._comment_obj:
             self.get()
@@ -325,8 +322,7 @@ class GitHubPRComment(BasePRComment):
                    tries=5, delay=1, backoff=2, max_delay=30)
 
     def append(self, text_to_append):
-        if not self.id:
-            raise Exception("'id' must be set to append to a comment.")
+        self._require_id()
         # Ensure comment object is present and up to date
         self.get()
         self.edit(self.body + text_to_append)
@@ -350,14 +346,12 @@ class GitLabPRComment(BasePRComment):
         return None
 
     def get(self):
-        if not self.id:
-            raise Exception("'id' must be set to get a comment.")
+        self._require_id()
         self._comment_obj = self._pr_obj.notes.get(self.id)
         self.body = self._comment_obj.body
 
     def create(self):
-        if not self.body:
-            raise Exception("'body' must be set to create a comment.")
+        self._require_body()
         if self.id:
             # Safeguard: id should not be set when body is initialized, but guard
             # against accidental duplicate comments if id was set after initialization.
@@ -367,8 +361,7 @@ class GitLabPRComment(BasePRComment):
             self.id = self._comment_obj.id
 
     def edit(self, new_body):
-        if not self.id:
-            raise Exception("'id' must be set to edit a comment.")
+        self._require_id()
         # Ensure comment object is present
         if not self._comment_obj:
             self.get()
@@ -377,8 +370,7 @@ class GitLabPRComment(BasePRComment):
         self._comment_obj.save()
 
     def append(self, text_to_append):
-        if not self.id:
-            raise Exception("'id' must be set to append to a comment.")
+        self._require_id()
         # Ensure comment object and body are present and up to date
         self.get()
         self.edit(self.body + text_to_append)
--- a/tools/pr_comments.py
+++ b/tools/pr_comments.py
@@ -276,8 +276,8 @@ class GitHubPRComment(BasePRComment):
         if not self.body:
             raise Exception("'body' must be set to create a comment.")
         if self.id:
-        # Safeguard: id should not be set when body is initialized, but guard
-        # against accidental duplicate comments if id was set after initialization.
+            # Safeguard: id should not be set when body is initialized, but guard
+            # against accidental duplicate comments if id was set after initialization.
             return
         self._comment_obj = retry_call(self._pr_obj.create_issue_comment, fargs=[self.body],
                                        exceptions=Exception, tries=3, delay=1, backoff=2, max_delay=10)
@@ -329,8 +329,8 @@ class GitLabPRComment(BasePRComment):
         if not self.body:
             raise Exception("'body' must be set to create a comment.")
         if self.id:
-        # Safeguard: id should not be set when body is initialized, but guard
-        # against accidental duplicate comments if id was set after initialization.
+            # Safeguard: id should not be set when body is initialized, but guard
+            # against accidental duplicate comments if id was set after initialization.
             return
         self._comment_obj = self._pr_obj.notes.create({"body": self.body})
         if self._comment_obj:
--- a/tools/pr_comments.py
+++ b/tools/pr_comments.py
@@ -46,7 +46,7 @@ class ChatLevels(Enum):
 
 def create_comment(repo_name, pr_number, comment, req_chatlevel):
     """
-    Create a comment to a pull request on GitHub
+    Create a comment to a pull request
 
     Args:
         repo_name (string): name of the repository
@@ -55,8 +55,7 @@ def create_comment(repo_name, pr_number, comment, req_chatlevel):
         req_chatlevel (member of ChatLevels Enum): minimum required chattiness level for creating the PR comment
 
     Returns:
-        github.IssueComment.IssueComment instance or None (note, github refers to
-            PyGithub, not the github from the internal connections module)
+        PRComment instance or None
     """
     fn = sys._getframe().f_code.co_name
 
@@ -65,12 +64,12 @@ def create_comment(repo_name, pr_number, comment, req_chatlevel):
         config.BOT_CONTROL_SETTING_CHATLEVEL, ChatLevels.BASIC.name).upper()
 
     if ChatLevels[chatlevel].value >= req_chatlevel.value:
-        gh = github.get_instance()
-        repo = gh.get_repo(repo_name)
-        pull_request = repo.get_pull(pr_number)
-        issue_comment = retry_call(pull_request.create_issue_comment, fargs=[comment],
-                                   exceptions=Exception, tries=3, delay=1, backoff=2, max_delay=10)
-        return issue_comment
+        pr_comment = create_pr_comment_instance(repo_name, pr_number, body=comment)
+        pr_comment.create()
+        # If 'id' is not set, something went wrong
+        if not pr_comment.id:
+            return None
+        return pr_comment
 
     else:
         log(f"{fn}(): not creating PR comment: "
--- a/tools/pr_comments.py
+++ b/tools/pr_comments.py
@@ -234,7 +234,7 @@ class BasePRComment(ABC):
     def body(self, value):
         # Ensure 'body' is not a blank string
         if value is not None and not value.strip():
-            raise ValueError("'body' must not be empty") 
+            raise ValueError("'body' must not be empty")
         self._body = value
 
     @property
@@ -245,7 +245,7 @@ class BasePRComment(ABC):
     def id(self, value):
         # Ensure 'id' is not a blank string
         if value is not None and not str(value).strip():
-            raise ValueError("'id' must not be empty") 
+            raise ValueError("'id' must not be empty")
         self._id = value
 
     def _require_body(self):
--- a/tools/pr_comments.py
+++ b/tools/pr_comments.py
@@ -30,7 +30,7 @@ from connections import github
 from tools import config
 
 
-PRComment = namedtuple('PRComment', ('repo_name', 'pr_number', 'pr_comment_id'))
+PRCommentInfo = namedtuple('PRCommentInfo', ('repo_name', 'pr_number', 'pr_comment_id'))
 
 
 class ChatLevels(Enum):
--- a/tools/pr_comments.py
+++ b/tools/pr_comments.py
@@ -276,7 +276,8 @@ class GitHubPRComment(BasePRComment):
         if not self.body:
             raise Exception("'body' must be set to create a comment.")
         if self.id:
-            # Return early if 'id' is set to avoid creating duplicate comments
+        # Safeguard: id should not be set when body is initialized, but guard
+        # against accidental duplicate comments if id was set after initialization.
             return
         self._comment_obj = retry_call(self._pr_obj.create_issue_comment, fargs=[self.body],
                                        exceptions=Exception, tries=3, delay=1, backoff=2, max_delay=10)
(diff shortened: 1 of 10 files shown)


tools/pr_comments.py
```diff
@@ -62,12 +64,12 @@ def create_comment(repo_name, pr_number, comment, req_chatlevel):
         config.BOT_CONTROL_SETTING_CHATLEVEL, ChatLevels.BASIC.name).upper()
 
     if ChatLevels[chatlevel].value >= req_chatlevel.value:
-        gh = github.get_instance()
-        repo = gh.get_repo(repo_name)
-        pull_request = repo.get_pull(pr_number)
-        issue_comment = retry_call(pull_request.create_issue_comment, fargs=[comment],
-                                   exceptions=Exception, tries=3, delay=1, backoff=2, max_delay=10)
-        return issue_comment
+        pr_comment = create_pr_comment_instance(repo_name, pr_number, body=comment)
```
trz42, line 68: The `retry_call` is no longer needed or errors are handled differently?

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+            err_msg = "Do not use this base class directly. "
+            err_msg += "Please use one of its subclasses instead."
+            raise NotImplementedError(err_msg)
+
+        # 'body' should be provided when creating a new comment
+        # 'id' should be provided when dealing with an existing comment
+        if (body and id) or not (body or id):
+            err_msg = "Exactly one of 'body' and 'id' must be "
+            err_msg += "set when initializing a comment class."
+            raise Exception(err_msg)
```
trz42: I'd modify this a bit:
- derive BasePRComment from `ABC` since it is an abstract class $\rightarrow$ the first `if` statement can be removed (maybe this could have been done for `EventInfo` too)
- expand the docstring with information about the arguments of the constructor
- clarify the comment for validating the arguments `body` and `id`
- also change the `if` logic to handle more edge cases (XOR looks good, but doesn't handle edge cases)
- make sure `id` is not an empty string
- use `ValueError` instead of a plain `Exception`

For example, add `from abc import ABC` and then refine the start of the class as follows:

```suggestion
class BasePRComment(ABC):
    """
    Base class to use for handling PR comments, which works differently for GitHub vs. GitLab.
    
    Args:
        repo_name (str): Name of the repository
        pr_number (int): Pull request number
        body (str, optional): Comment body. Provide when creating a new comment.
        id (int, optional): ID of an existing comment. Provide when referencing an existing comment.
        
    Exactly one of body or id must be provided.
    """
    def __init__(self, repo_name, pr_number, body=None, id=None):
        # Exactly one of body (new comment) or id (existing comment) must be provided.
        if (body is None) == (id is None):
            err_msg = "Exactly one of 'body' or 'id' must be "
            err_msg += "set when initializing a comment class."
            raise ValueError("Exactly one of 'body' and 'id' must be set when initializing a comment class.")

        if id is not None and not str(id).strip():
            raise ValueError("'id' must not be empty")
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        self.body = body
+        self.id = id
+        self.repo_name = repo_name
+        self.pr_number = pr_number
+        self._pr_obj = None
+        self._comment_obj = None
+
+    @property
+    def html_url(self):
+        raise NotImplementedError()
```
trz42: Make method abstract by prepending `@abstractmethod` and just let it `pass`


```suggestion
    @property
    @abstractmethod
    def html_url(self):
        pass
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+
+    def create(self):
+        raise NotImplementedError()
+
+    def edit(self):
+        raise NotImplementedError()
+
+    def append(self):
+        raise NotImplementedError()
```
trz42: Make methods abstract and add arguments for `edit` and `append`, for example,


```suggestion
    @abstractmethod
    def get(self):
        pass

    @abstractmethod
    def create(self):
        pass

    @abstractmethod
    def edit(self, new_body):
        pass

    @abstractmethod
    def append(self, text_to_append):
        pass
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+
```
trz42, line 279: I'd add two non-abstract helper functions which simplify checks in the derived classes:

```python
    def _require_id(self):
        if not self.id:
            raise ValueError("'id' must be set.")

    def _require_body(self):
        if self.body is None:
            raise ValueError("'body' must be set.")
```

Note, these are not defined with `@abstractmethod` so can be used directly and only require an implementation in the derived classes if different functionality is needed.

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+    @property
+    def html_url(self):
+        if self._comment_obj:
+            return self._comment_obj.html_url
+        return None
+
+    def get(self):
+        if not self.id:
+            raise Exception("'id' must be set to get a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_id()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+            # Return early if 'id' is set to avoid creating duplicate comments
+            return
+        self._comment_obj = retry_call(self._pr_obj.create_issue_comment, fargs=[self.body],
+                                       exceptions=Exception, tries=3, delay=1, backoff=2, max_delay=10)
+        if self._comment_obj:
+            self.id = self._comment_obj.id
+
+    def edit(self, new_body):
+        if not self.id:
+            raise Exception("'id' must be set to edit a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_id()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        # Ensure comment object is present
+        if not self._comment_obj:
+            self.get()
+        self.body = new_body
+        retry_call(self._comment_obj.edit, fargs=[self.body], exceptions=Exception,
+                   tries=5, delay=1, backoff=2, max_delay=30)
+
+    def append(self, text_to_append):
+        if not self.id:
+            raise Exception("'id' must be set to append to a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_id()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+    @property
+    def html_url(self):
+        if self._comment_obj:
+            # GitLab comment object does not include a comment URL
+            return f"{self._pr_obj.web_url}#note_{self._comment_obj.id}"
+        return None
+
+    def get(self):
+        if not self.id:
+            raise Exception("'id' must be set to get a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_id()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        if self.id:
+            # Return early if 'id' is set to avoid creating duplicate comments
+            return
+        self._comment_obj = self._pr_obj.notes.create({"body": self.body})
+        if self._comment_obj:
+            self.id = self._comment_obj.id
+
+    def edit(self, new_body):
+        if not self.id:
+            raise Exception("'id' must be set to edit a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_id()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        # Ensure comment object is present
+        if not self._comment_obj:
+            self.get()
+        self.body = new_body
+        self._comment_obj.body = self.body
+        self._comment_obj.save()
+
+    def append(self, text_to_append):
+        if not self.id:
+            raise Exception("'id' must be set to append to a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_id()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+
+    def get(self):
+        if not self.id:
+            raise Exception("'id' must be set to get a comment.")
+        self._comment_obj = self._pr_obj.notes.get(self.id)
+        self.body = self._comment_obj.body
+
+    def create(self):
+        if not self.body:
+            raise Exception("'body' must be set to create a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_body()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        if not self.id:
+            raise Exception("'id' must be set to get a comment.")
+        self._comment_obj = retry_call(self._pr_obj.get_issue_comment, fargs=[self.id],
+                                       exceptions=Exception, tries=5, delay=1, backoff=2, max_delay=30)
+        if self._comment_obj:
+            self.body = self._comment_obj.body
+
+    def create(self):
+        if not self.body:
+            raise Exception("'body' must be set to create a comment.")
```
trz42: Use helper function:

```suggestion
        self._require_body()
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        if not self.id:
+            raise Exception("'id' must be set to get a comment.")
+        self._comment_obj = self._pr_obj.notes.get(self.id)
+        self.body = self._comment_obj.body
+
+    def create(self):
+        if not self.body:
+            raise Exception("'body' must be set to create a comment.")
+        if self.id:
+            # Return early if 'id' is set to avoid creating duplicate comments
```
trz42: Maybe revise this comment a little, for example,

```suggestion
        # Safeguard: id should not be set when body is initialized, but guard
        # against accidental duplicate comments if id was set after initialization.
```

tools/pr_comments.py
```diff
@@ -192,3 +194,175 @@ def update_pr_comment(event_info, update):
...
+        self._comment_obj = retry_call(self._pr_obj.get_issue_comment, fargs=[self.id],
+                                       exceptions=Exception, tries=5, delay=1, backoff=2, max_delay=30)
+        if self._comment_obj:
+            self.body = self._comment_obj.body
+
+    def create(self):
+        if not self.body:
+            raise Exception("'body' must be set to create a comment.")
+        if self.id:
+            # Return early if 'id' is set to avoid creating duplicate comments
```
trz42: Maybe revise this comment a little, for example,

```suggestion
        # Safeguard: id should not be set when body is initialized, but guard
        # against accidental duplicate comments if id was set after initialization.
```

tools/pr_comments.py
trz42: Curious, would it make sense to allow an empty string for `self.body`?

trz42 (changes requested):
Thanks a lot for this PR @sondrebr ! Looks quite good. A few comments added for consideration.

I'll try to run it too, to get a feeling what is needed to setup a bot responding to web hooks from GitLab.

tools/pr_comments.py
sondrebr: Good point, I don't think GH or GL accept any kind of blank comment. Might be worth making `body` and `id` properties and verifying the values in the setters, maybe like this?
```python
    def _require_body(self):
        if self.body is None:
            raise ValueError("'body' must be set.")

    def _require_id(self):
        if self.id is None:
            raise ValueError("'id' must be set.")

    @property
    def body(self):
        return self._body

    @body.setter
    def body(self, value):
        if value is not None and not value.strip():
            raise ValueError("'body' must not be empty") 
        self._body = value

    @property
    def id(self):
        return self._id

    @id.setter
    def id(self, value):
        if value is not None and not str(value).strip():
            raise ValueError("'id' must not be empty") 
        self._id = value
```

tools/pr_comments.py
sondrebr: Retries are handled by the PRComment classes now - GitHubPRComment methods use `retry_call` to handle retries, while retries in the GitLabPRComment methods are handled by the `python-gitlab` module.

tools/pr_comments.py
sondrebr: Used the suggestion with minimal changes in da996a17dbab73f9f4583e59fb76d9d88d5384c1 👍

tools/pr_comments.py
sondrebr: Done: da996a17dbab73f9f4583e59fb76d9d88d5384c1

tools/pr_comments.py
sondrebr: Done: da996a17dbab73f9f4583e59fb76d9d88d5384c1

tools/pr_comments.py
```diff
@@ -192,3 +195,187 @@ def update_pr_comment(event_info, update):
...
+        self._comment_obj = retry_call(self._pr_obj.get_issue_comment, fargs=[self.id],
+                                       exceptions=Exception, tries=5, delay=1, backoff=2, max_delay=30)
+        if self._comment_obj:
+            self.body = self._comment_obj.body
+
+    def create(self):
+        if not self.body:
+            raise Exception("'body' must be set to create a comment.")
+        if self.id:
+        # Safeguard: id should not be set when body is initialized, but guard
```
sondrebr: 😝

Edit: Hound noticed the missing indentation

tools/pr_comments.py
trz42: Good idea!

sondrebr: Added in 7a25f23251e334b874150f73bb4a21ed433db661 and 885226735cc6e3841c8ca0c6527e9a3a91f510da

trz42 (commented):
Looks good to me! I'll give it another try against GitLab and also check if it works on GitHub, and merge it then.

trz42:
> Looks good to me! I'll give it another try against GitLab and also check if it works on GitHub, and merge it then.

Tested PR against both GitLab and GitHub with all supported commands. All good.

trz42 (approved):
Thanks a lot for this @sondrebr !

Issue #375, opened 2026-05-18
add unit tests covering GitLab support


trz42:
We need to extend unit test coverage for the recent developments to support GitLab repositories provided by the PRs:

- #370

PR #379, opened 2026-06-24
[gitlab] Improve test coverage

sondrebr:
This PR focuses on improving test coverage for the stuff added in the previous GitLab PRs (#370 and #376). It contains:
- Tests for
  - `connections/gitlab.py`
  - `tools/git.py`
  - `tools/commands.py` (`SUPPORTED_COMMANDS_PER_GIT_HOST` and `get_supported_commands()`)
  - `tools/event_info.py`
- Sample webhook events for use in tests (stored as JSON files in `test/events/`)
- Some fixes and tiny changes

The sample event files are based on webhook events from GitHub and GitLab, with some values changed.
I did however keep (I think) all the fields so they can also be used for reference and tests in the future, for example when adding new EventInfo properties.

Note that this PR does not include any tests for the PRComment classes.
Existing code in `tools/pr_comments.py` will probably have to be changed a lot when implementing GL support for the rest of the commands, so I think it's better to wait and see if the existing tests can be modified and reused for PRComment tests.

Edit: Addresses issue #375


tests/test_connections_gitlab.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def auth(self):
+        # Set 'user' when calling auth()
+        self.user = self._objects.CurrentUser()
+
+
+# Test verify_connection()
+def test_verify_connection(capfd):
+    # Token is valid and current user is successfully retrieved and stored
+    mock_gl = MockGitlab(private_token=ACCESS_TOKEN)
```
trz42: The argument is only needed so no `ValueError` is raised, it is not verified in the following line? If so, this should be made more clear, e.g.,

```suggestion
    mock_gl = MockGitlab(private_token="irrelevant-for-next-line-but-needed-or-constructor-would-raise-ValueError")
```

tests/test_connections_gitlab.py
```diff
@@ -0,0 +1,145 @@
...
+    gitlab.verify_connection(mock_gl)
```
trz42: Given that no URL was provided when creating `mock_gl` one could assume that `verify_connection` might actually make a call to `gitlab.com`. So here a comment such as the following would make this easier to understand

```python
# verify_connection() calls gl.auth() and checks gl.user;
# MockGitlab.auth() and thus mock_gl.auth() sets user without any network I/O.
```

tests/test_connections_gitlab.py
```diff
@@ -0,0 +1,145 @@
...
+        # Error message should mention the 'user' attribute and the 'CurrentUser' type
+        err_msg = capfd.readouterr().err
+        assert "'user'" in err_msg
+        assert "'CurrentUser'" in err_msg
+
+
+# Test connect()
+@patch("connections.gitlab.gitlab.Gitlab", MockGitlab)
+def test_connect(capfd):
+    # Test successful connection
```
trz42: This comment is misleading because no connection is actually established. What it should say is why the environment variable named via `PAT_ENV_VAR_NAME` is set (because `gitlab.connect()` needs it). Example comment:

```suggestion
    # connect() is expected to read the access token from the environment
    # variable $GITLAB_PROJECT_ACCESS_TOKEN (defined via constant PAT_ENV_VAR_NAME)
```

tests/test_connections_gitlab.py
```diff
@@ -0,0 +1,145 @@
...
+    os.environ[PAT_ENV_VAR_NAME] = ACCESS_TOKEN
+    with patch("connections.gitlab.verify_connection") as mock_verify_connection:
+        gitlab.connect()
+        # connect() should verify connection
```
trz42: No connection is actually verified by the mocked `verify_connection`. A more clear comment could help the reader to understand the assert. For example,

```suggestion
        # verify that connect() called verify_connection()
```

tests/test_connections_gitlab.py
```diff
@@ -0,0 +1,145 @@
...
+        mock_verify_connection.assert_called()
+    # Created connection should be stored as '_gl' - verify properties
```
trz42: While not wrong, this comment misses an opportunity to explain how different parts come together and that is verified by the following asserts. Particularly, connect() calls
```
_gl = gitlab.Gitlab(url, access_token, timeout=timeout, retry_transient_errors=True)
```
where `url` and `timeout` are obtained from the bot's config and `access_token` via the bespoke env var. `MockGitlab` just uses all values to set corresponding entries in the created instance. We ensure that connect() creates the correct instance and values with the following asserts.

Maybe something like the following could explain this in "a bit" more detail (maybe a bit too much tough):

```suggestion
    # connect() creates the client via
    #   gitlab.Gitlab(url, access_token, timeout=timeout, retry_transient_errors=True)
    # with 'url' and 'timeout' from the bot's config and 'access_token' from
    # PAT_ENV_VAR_NAME, and stores it as module-level '_gl'.
    # Since Gitlab is patched with MockGitlab, which simply stores all constructor
    # arguments as attributes, the asserts below verify that connect() passed the
    # expected values.
```

connections/gitlab.py
```diff
@@ -63,9 +63,7 @@ def connect():
     else:
         del os.environ['GITLAB_PROJECT_ACCESS_TOKEN']
```
trz42, line 64: While not changed here, I wonder if this prevents the bot from ever calling this function again? Otherwise stated, any second call would crash the bot due to `logging.error` in case the env var is not set.

tests/test_connections_gitlab.py
trz42: While not wrong, this comment misses an opportunity to explain how different parts come together and that is verified by the following asserts. Particularly, connect() calls
```
_gl = gitlab.Gitlab(url, access_token, timeout=timeout, retry_transient_errors=True)
```
where `url` and `timeout` are obtained from the bot's config and `access_token` via the bespoke env var. `MockGitlab` just uses all values to set corresponding entries in the created instance. We ensure that connect() creates the correct instance and values with the following asserts.

tests/test_connections_gitlab.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Test with missing environment variable 'GITLAB_PROJECT_ACCESS_TOKEN' - should exit
+    with pytest.raises(SystemExit):
+        gitlab.connect()
+    # Error message should mention the environment variable
+    assert PAT_ENV_VAR_NAME in capfd.readouterr().err
+
+
+# Test get_instance()
+@patch("connections.gitlab.gitlab.Gitlab", MockGitlab)
```
trz42: Is this global patch needed?

tests/test_tools_commands.py
```diff
@@ -11,12 +11,52 @@
...
+# Test SUPPORTED_COMMANDS_PER_GIT_HOST
+@pytest.mark.parametrize("git_host", git.SUPPORTED_GIT_HOSTS)
+def test_support_commands_per_git_host(git_host):
+    # There should be an entry for each supported Git hosting platform
+    assert git_host in commands.SUPPORTED_COMMANDS_PER_GIT_HOST
+
+    # Each entry should be a list of commands
+    supported_commands = commands.SUPPORTED_COMMANDS_PER_GIT_HOST[git_host]
+    assert isinstance(supported_commands, list)
+    assert all([command in commands.ALL_COMMANDS for command in supported_commands])
```
trz42: The second assert seems to verify something else than "Each entry should be a list of commands"

tests/test_tools_git.py
```diff
@@ -0,0 +1,136 @@
...
+    mock_read_config.assert_not_called()
+
+    git._git_host = None
+
+    # Test without provided cfg
+    with expectation as expected:
+        assert git.get_git_hosting_platform() == expected
+    mock_read_config.assert_called_once()
+
+    git._git_host = None
```
trz42: While not wrong, I think the test becomes easier to understand and maintain if we split testing valid configs and invalid configs into two tests. For example,

```suggestion
# Test get_git_hosting_platform() - valid configs
@pytest.mark.parametrize("cfg,expected", [
    # 'hosting_platform' set to 'github'
    (CFG, git.GITHUB),
    (GITHUB_CFG, git.GITHUB),

    # 'hosting_platform' set to 'gitlab'
    (GITLAB_CFG, git.GITLAB),
])
@patch("tools.config.read_config")
def test_get_git_hosting_platform_valid(mock_read_config, cfg, expected):
    mock_read_config.return_value = cfg

    # Test with provided cfg
    git._git_host = None
    assert git.get_git_hosting_platform(cfg) == expected
    mock_read_config.assert_not_called()

    # Test without provided cfg
    git._git_host = None
    assert git.get_git_hosting_platform() == expected
    mock_read_config.assert_called_once()

    git._git_host = None


# Test get_git_hosting_platform() - invalid configs, should exit
@pytest.mark.parametrize("cfg", [
    # 'hosting_platform' set to an invalid value
    UNSUPPORTED_PLATFORM_CFG,

    # 'hosting_platform' not set
    NO_HOSTING_PLATFORM_CFG,
])
@patch("tools.config.read_config")
def test_get_git_hosting_platform_invalid(mock_read_config, cfg):
    mock_read_config.return_value = cfg

    # Test with provided cfg
    git._git_host = None
    with pytest.raises(SystemExit):
        git.get_git_hosting_platform(cfg)
    mock_read_config.assert_not_called()

    # Test without provided cfg
    git._git_host = None
    with pytest.raises(SystemExit):
        git.get_git_hosting_platform()
    mock_read_config.assert_called_once()

    git._git_host = None
```

tests/test_tools_event_info.py
```diff
@@ -0,0 +1,371 @@
...
+
+    # Comment events
+    COMMENT_CREATED: "test/events/gitlab/comment-created.json",
+    COMMENT_EDITED: "test/events/gitlab/comment-edited.json",
+}
+
+
+# PyGHee's get_event_info() expects a flask.Request object as input
+# This class roughly implements the attributes used by the PyGHee function
+# 'event' passed on initialization must be a dict read from one of the sample event files
```
trz42: Maybe this could be more easily understood if it follows common docstring format?

For example,

```python
class MockRequest():
    """Mock of a flask.Request object as expected by PyGHee's get_event_info().

    Only the attributes used by that function are implemented.
    """

    def __init__(self, event):
        """Set up mock request attributes from a sample event.

        Args:
            event: dict read from one of the sample event files;
                must contain the keys 'json' and 'headers'.
        """
        # remainder of __init__
```

tests/test_tools_event_info.py
```diff
@@ -0,0 +1,371 @@
...
+class MockRequest():
+    def __init__(self, event):
+        self.json = event["json"]
+        self.data = json.dumps(self.json).encode()
+        self.headers = CaseInsensitiveDict(event["headers"])
+
+
+# Mock class handling individual user information in MockGitlab
+class MockGLUser():
```
trz42: I think `MockGitlabUser` would be more readable. Keep as is if the mocked class is `GLUser`.

tests/test_tools_event_info.py
```diff
@@ -0,0 +1,371 @@
...
+    def __init__(self, id, username):
+        self.id = id
+        self.username = username
+
+
+# Mock class handling collections of users in MockGitlab
+class MockGLUsers():
```
trz42: Better use `MockGitlabUsers`. Keep as is if the mocked class is `GLUsers`.

tests/test_tools_event_info.py
```diff
@@ -0,0 +1,371 @@
...
+    def __init__(self, users):
+        # Store ID -> User mappings
+        self._users = {id: MockGLUser(id, username) for id, username in users.items()}
+
+    def get(self, id, *args, **kwargs):
```
trz42: Since `args` and `kwargs` aren't used we should either add a comment that explains why they are listed or use specific names that make sure linters won't flag them as an issue, e.g., adding `_` as prefix or even use `*_` and `*__`.

tests/test_tools_event_info.py
```diff
@@ -0,0 +1,371 @@
...
+    def __init__(self, users=None):
+        self.users = MockGLUsers(users)
+
+
+# Read event from file, create a MockRequest and return event_info dict from PyGHee's get_event_info()
+def get_event_info_from_file(path):
+    # Use get_git_hosting_platform() imported by event_info - return value should be patched by tests
+    event_source = event_info.get_git_hosting_platform()
+    with open(path, "r") as f:
+        event = json.load(f)
```
trz42: minor comment, please don't use single letter variable names

```suggestion
    with open(path, "r") as file:
        event = json.load(file)
```

trz42 (changes requested):
Very nice PR! A few suggestions mostly for clarifying comments or docstrings. Some code seems unnecessary (one `@patch` decorator). One test function might be better split in two (`test_get_git_hosting_platform`).

connections/gitlab.py
sondrebr: Correct. Maybe we should add a check at the start of `connect()` in another PR? E.g.,
```Python
# (add to docstring): Should only be called once. Does nothing after initial call.
if _gl is not None:
    logging.log("connect() has already been called. Returning.")
    return
```

tests/test_connections_gitlab.py
sondrebr: I added it as `irrevelant-for-auth-...` (along with the comments before `verify_connection()`), hopefully clearer than just using `ACCESS_TOKEN`:
4d6069bfc6263774883ed6e294b78c24b8e5a585

tests/test_connections_gitlab.py
sondrebr: Added in 4d6069bfc6263774883ed6e294b78c24b8e5a585 👍

tests/test_connections_gitlab.py
sondrebr: Fair point. Fixed in 4d6069bfc6263774883ed6e294b78c24b8e5a585

tests/test_connections_gitlab.py
sondrebr: Fixed in 4d6069bfc6263774883ed6e294b78c24b8e5a585

tests/test_connections_gitlab.py
sondrebr: A more detailed explanation might be useful, yes. Added your suggestion with minor changes in 4d6069bfc6263774883ed6e294b78c24b8e5a585

tests/test_connections_gitlab.py
sondrebr: Nope, must've been left there by accident 😶 Removed in 4d6069bfc6263774883ed6e294b78c24b8e5a585

tests/test_tools_event_info.py
sondrebr: Good suggestion. Added in afff4612bec5f85bc3ece8d6bb8b21974fbb5d08

tests/test_tools_event_info.py
sondrebr: Renamed to `MockGitlabUser` in afff4612bec5f85bc3ece8d6bb8b21974fbb5d08

tests/test_tools_event_info.py
sondrebr: Renamed to `MockGitlabUsers` in afff4612bec5f85bc3ece8d6bb8b21974fbb5d08

tests/test_tools_event_info.py
sondrebr: I renamed them to `_args` and `_kwargs` to indicate that they are unused. Made pylint happy at least 👍 afff4612bec5f85bc3ece8d6bb8b21974fbb5d08

tests/test_tools_event_info.py
sondrebr: Sorry, bad habit 😅 Fixed in afff4612bec5f85bc3ece8d6bb8b21974fbb5d08

tests/test_tools_git.py
sondrebr: Done: 9672f4307c143bcf421d79c15eee9ae1ec379fec

tests/test_tools_commands.py
sondrebr: I can see how this might be confusing. c67ac49728357f35e5f7582720e903cdf0c68bb5 hopefully makes it clearer and more readable.

connections/gitlab.py
trz42: Sounds good to me. Could you open a issue, so we don't forget this?

trz42 (approved):
Thanks for the update @sondrebr ! Looks good to me.

Nice if you could open a new issue for what is discussed in https://github.com/EESSI/eessi-bot-software-layer/pull/379#discussion_r3517776825

connections/gitlab.py
sondrebr: OK: #385

sondrebr:
> Nice if you could open a new issue for what is discussed in [#379 (comment)](https://github.com/EESSI/eessi-bot-software-layer/pull/379#discussion_r3517776825)

Done 👍 #385

PR #384, opened 2026-07-21
Consolidate repos config

casparvl:
In the bot-configs repo, we currently have a lot of duplication of the repos configuration. It makes sense to consolidate that, and actually put it in a public place - there are no secrets here, just CVMFS configuration.

It is slightly different between for the dev.eessi.io stuff. I'm wondering if it's worth just consolidating all of that in a single repos.cfg - but that can be discussed during the review (it shouldn't hurt to have _extra_ repositories defined in a repos.cfg, I guess).

This has companion PR in `bot-configs`: https://github.com/EESSI/bot-configs/pull/98

Disclaimer: work was done by an AI agent, reviewed by me.


casparvl:
Discussed with @bedroge , we'll do this in a separate repo as we don't want to have to update our bot version just to get new repo configuration deployed.

